PR #2, opened 2015-06-01
Addresses

tie-rack:
Pivotal: [95494362](https://www.pivotaltracker.com/story/show/95494362)

Changes the `:user/address` attribute to cardinality many in the Datomic schema and adds an :address/type attribute, allowing for multiple, arbitrarily-typed addresses for each user.

For example:

```clojure
{:first-name "John"
 :last-name "Adams"
 :email "jadams1735@example.com"
 :phone "202-456-1414"
 :addresses {:home {:street "1600 Pennsylvania Ave NW"
                    :city "Washington"
                    :state "DC"
                    :zip "20500"}
             :birthplace {:street "133 Franklin St"
                          :city "Quincy"
                          :state "MA"
                          :zip "02169"}}}
```

Includes namespace gardening so everything's not all sitting in `user-works.handlers`.


README.md
```diff
@@ -74,10 +76,10 @@ The response will include your changes:
         :first-name "Jack"
         :last-name "Adams"
         :email "jadams1735@example.com"
-        :address {:street "1200 Pennsylvania Ave NW"
-                  :city "Washington"
-                  :state "DC"
-                  :zip "20500"}}
+        :addresses ({:street "1200 Pennsylvania Ave NW"
```
tank157: The response here will include type too, no?

src/user_works/entities/address.clj
```diff
@@ -0,0 +1,68 @@
+(ns user-works.entities.address
```
tank157: Out of scope for right now, but at some point this could/should be generified so we can reuse the base code for addresses attached to other things too (like election authorities).

tank157:
LGTM!

PR #4, opened 2015-06-09
Allow core.async tests to timeout

cap10morgan:
...as suggested by @ericnormand.

I tested that the tests pass and that they fail in the way we expect when I comment out the channel puts in the `handlers` ns.

Assigning @tie-rack since he looked at the equivalent PR in election-works.


tie-rack:
This one also LGTM.

PR #5, opened 2015-06-11
Search user by ocd

ericnormand:
### Context

Users have addresses and addresses can have ocd-ids (according to the schema). Addresses have a type, among which is :registered.

The task here is to make a new message queue that allows for searching for users by a single ocd-id that matches one on their :registered address. It's a straightforward datomic query. Note that it only performs an exact string match.

[Pivotal Card](https://www.pivotaltracker.com/story/show/95877904)

### How to use

Send a message to `user-works.user.search` with the following format:

```
{:district-division "<ocd-id>"}
```

### Validation

There's validation for the message format. The ocd-id is checked against a regex we also use in election-works.

### About the tests

There are tests for several cases, including error cases. The one funny part is that at the time of branching, there was no geocoding of addresses built into user-works. That means that when you create a user, there's no ocd-ids associated with them. So in the tests we manually transact ocd-ids onto them. **These tests will likely need to be revisited one actual ocd-ids are added.**


ericnormand:
Tests passed using `lein test`.

Manually tested queue wiring. :+1:

ericnormand:
Assigning to the legendary @tie-rack.

project.clj
```diff
@@ -21,5 +22,5 @@
                                    :password [:gpg :env]}}
   :uberjar-name "user-works.jar"
   :profiles {:uberjar {:aot :all}
-             :dev {:resource-paths ["dev-resources"]}
+             :dev {:resource-paths ["dev-resources" "test-resources"]}
```
tie-rack: Why is the dev profile using test resources?

ericnormand: Because running tests from the cider repl needs test resources.

tie-rack: Overwriting dev-resources with test-resources is problematic. You can connect to a repl with the test profile without changing the dev profile.

tie-rack: If you call `cider-jack-in` with a prefix argument, you can change the command used to start the repl. Thus, `C-u C-c M-j` will prompt you with the arguments to `lein repl`, at which point you can add `with-profile +test`.

The dev profile is used, too, in `lein run` where test-resources may interfere with the expected behavior of running the project.

tie-rack:
LGTM!

PR #6, opened 2015-06-11
Geocode addresses

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/95983100)

We exposed sending in ocd-division-ids to the create and update, and if provided those values will be honored. If not provided, we'll make a call to address-works and incorporate the ocd-divisions into the addresses.

Future issues to consider:
* What's the full sequence of things we should do on an address-works call timeout? It's currently going to wait up to 5 seconds before giving up, logging an error, and moving on (which means just not including ocd-ids, but saving regardless)
* Should we look at any other data from address-works? Should we incorporated enhanced address data, changes from postal standardization,etc?
* Should we do anything special if no ocds come back?
* Should we restrict the lookup to a single address type, and if so, what is that address type, and how do we codify that special type? Right now type is pretty much a freeform keyword.

Assigning to @ericnormand


src/user_works/entities/address.clj
```diff
@@ -49,15 +52,25 @@
...
+                 current-v (get current-address db-k)
+                 new-v (get new-address k)]
+           :when (not= current-v new-v)]
+       (if new-v
+         [:db/add address-id db-k new-v]
+         [:db/retract address-id db-k current-v]))
+     (for [ocd-id new-ocd-ids]
+       [:db/add address-id ocd-id-db-k ocd-id])
+     (for [ocd-id old-ocd-ids]
+       [:db/retract address-id ocd-id-db-k ocd-id]))))
```
ericnormand: I'm concerned about the lack of atomicity in our transactions. What do you say we make this into a datomic transaction function?

src/user_works/entities/address.clj
```diff
@@ -49,15 +52,25 @@
...
-          :when (not= current-v new-v)]
-      (if new-v
-        [:db/add address-id db-k new-v]
-        [:db/retract address-id db-k current-v]))))
+        address-id (:db/id current-address)
+        new-ocd-ids (set/difference (:ocd-division-id new-address)
+                                    (:address/ocd-division-id current-address))
+        old-ocd-ids (set/difference (:address/ocd-division-id current-address)
+                                    (:ocd-division-id new-address))
+        ocd-id-db-k (util/add-namespace "address" :ocd-division-id)]
```
ericnormand: What do you think about using the literal keyword `:address/ocd-division-id`. It's much clearer.

src/user_works/entities/address.clj
tie-rack: I don't understand. The atomicity of transactions is guaranteed by datomic. You can't make a non-atomic transaction.

src/user_works/handlers.clj
```diff
@@ -2,7 +2,34 @@
...
+      (if-not (= :timeout response)
+        (if-let [ocd-divisions (:ocd-divisions response)]
+          (merge address {:ocd-division-id ocd-divisions})
+          address)
+        (do (log/error "Response timeout calling address-works to geocode address: " address)
+            address)))))
+
+(defn geocode-addresses
+  [message]
+  (loop [keys (-> message :addresses keys vec)
```
ericnormand: What do you say to getting rid of this loop? Looks like it's mostly a `map-vals`.

ericnormand:
I'm a little shaky about how this is wired up. Who's listening on the `address` queue?

src/user_works/entities/address.clj
ericnormand: Not if you are basing the transaction data on an old database value.

```
Read entity ---------> calculate new data  ----------> write data to db
                     ^--------------------------------------^
                           this is where someone else changes the data while you're not looking!
```

+1 for ascii art!

ericnormand:
This is some nice work!

I've addressed a few issues in the code to get some discussion started. The two blockers are who is actually doing the geocoding and the atomicity of the transaction.

tank157:
So the "who is doing the geocoding" is mentioned in the PR and also is more or less shared tribal knowledge, but both of those are ephemeral so I'll add a comment or doc string to the code to address that.

As for the atomicity... that I'll have to look into some more and see if I understand Eric's concerns and if there's anything we can do in a system that makes asynchronous calls during the create and update operations.

ericnormand:
Oh, I see. It is mentioned in the PR. "call to address-works" is the same as the "addresses" queue.

ericnormand:
Now it's just the atomicity question blocking.

I'm open for a chat on Slack/ScreenHero if you'd like to discuss it.

tank157:
It is, but you bring up a good point that this is mostly shared tribal knowledge at this point, I'm going to at least add a comment indicating who is expected to be at the other end of that queue.

ericnormand:
Nice changes.

ericnormand:
Ok, atomicity is moved to two new cards. Thanks @tank157 

LGTM!

PR #7, opened 2015-06-18
[TH] change template from fake var to actual env vars

tank157:
No Pivotal Card, just getting it ready for deployment. I have the Buildkite project created, and filtered so it won't actually deploy anything yet until we are ready.

Assigning to @ericnormand


ericnormand:
LGTM

PR #8, opened 2015-06-22
Misc cleanups

cap10morgan:
Misc. cleanups and fixes made while I was reviewing a card.


ericnormand:
LGTM

cap10morgan:
@tie-rack You weren't loving the `:birthplace` -> `:registered` address change in the README. Care to elaborate why not?

tie-rack:
The data in the README moves away from historical accurateness (those really are the birthplaces of those people).

What I really object to is the fact that `:registered` addresses are a special case. The fact that users cannot be found _by id_ without having a `:registered` address is, frankly, bizarre for something that should be agnostic about addresses altogether. user-works's `find-by-id` should find the user (with _whatever_ data it has) no matter what kind of address the user has.

cap10morgan:
@tie-rack I agree. We should fix this. I'll write a card for adding the ability to restrict searches to an arbitrary address type when searching users by OCD-ID (which we will need because we don't want to match users on non-registered addresses when looking them up for electorate records, for example). That should make them non-special again.

In the meantime, I'll revert that change in here since it will hopefully soon provide no value for its historical inaccuracies.

PR #9, opened 2015-06-23
[TH] change :ocd-division-id to :ocd-divisions

tank157:
Requested change from [this Pivotal Card](https://www.pivotaltracker.com/story/show/95983100).

Assigning to requester, @cap10morgan


src/user_works/entities/address.clj
```diff
@@ -1,17 +1,25 @@
...
             [datomic-toolbox :as db]
-            [user-works.util :as util]))
+            [user-works.util :as util]
+            [clojure.set :as set]))
 
 (def division-regex #"\Aocd\-division\/country\:[a-z]{2}(?:\/[-\w]+\:[-\w.~]+)*\z")
 
 (def address-keys #{:type :street :street-2 :city :state :zip :ocd-division-id})
+(def singular-to-plural {:ocd-division-id :ocd-divisions})
+(def plural-to-singular {:ocd-divisions :ocd-division-id})
```
cap10morgan: This is more about internal vs. external representations, no (rather than singular vs. plural)?

src/user_works/entities/address.clj
```diff
@@ -1,17 +1,25 @@
...
+  (-> address
+      (assoc :type type)
+      (set/rename-keys plural-to-singular)
+      (util/namespace-map "address" address-keys)))
 
 (defn denamespace-address [address]
-  (util/denamespace-map address address-keys))
+  (-> address
+      (util/denamespace-map address-keys)
+      (set/rename-keys singular-to-plural)))
```
cap10morgan: Now it seems like these 2 fns are doing more than they claim to in their names. What about `->internal-representation` and `->external-representation`?

cap10morgan:
LGTM

PR #10, opened 2015-07-07
[TH DG] Return the user from a user create message.

tank157:
We needed access to some of the data inside the created user, and since everywhere else we tend to respond back with the whole entity we just created, we made the change to do the same here.

Assigning to @ericnormand 

Part of the end-to-end testing work [here](https://www.pivotaltracker.com/story/show/97637904)


cap10morgan:
FWIW, this LGTM

ericnormand:
I like it.

LGTM

PR #14, opened 2015-08-26
Require phone or email

tank157:
So I went above the card to do all create/update validations, because we had none. But it also incorporates the email and/or phone work.

[Pivotal Card](https://www.pivotaltracker.com/story/show/100896956)

Assigning @danielglauser


danielglauser:
LGTM

PR #18, opened 2015-09-15
Handle errors

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/102522276)

Pretty simple refactor, tested in krakenstein, but I haven't tested it directly with user-http-api at a curl level, so I'm going to start that now. But please go over the code.

Assigning to @danielglauser


src/user_works/handlers.clj
```diff
@@ -13,6 +13,26 @@
 (def augment-address (wire-up/async->fn channels/addresses))
 (def address-works-timeout 10000)
 
+(defn handle-errors
+  [handler]
+  (fn [message]
+    (try
+      (handler message)
+      (catch IllegalArgumentException semantic-error
+        {:status :error
```
danielglauser: Not sure what `:error` results in but shouldn't this return an HTTP 400 Bad Request?

cap10morgan: @danielglauser No, this is a -works component. It does not speak HTTP.

The user-http-api gateway should turn this into a 400 when it goes back to that.

tank157:
I ran through creating, reading, and deleting a user through user-http-api with these changes, and tried out some malformed messages, seemed good.

danielglauser:
LGTM

tank157:
Ok, this has been tested again with the latest master merged in and other projects updated to master in krakenstein, I'm going to merge it. It shouldn't have sat unmerged this long, my bad.

PR #19, opened 2015-09-30
Race dob citizenship

7maples:
[Pivotal Card](https://www.pivotaltracker.com/story/show/103374374)

Tested in krakenstein with success


danielglauser:
Should we have a set of races we allow? Or do you think that isn't necessary?

7maples:
@danielglauser that will be handled by the view under https://www.pivotaltracker.com/story/show/103463914

danielglauser:
Excellent. LGTM

PR #20, opened 2015-10-16
Add a test which creates a user with :mailing

ericnormand:
https://www.pivotaltracker.com/story/show/103374528

We want to make sure that we can pass in all types of addresses,
including a mailing address. I duplicated the basic user creation test
and added a mailing address along with the registered address. It passed
the first time. I also made sure it was being passed in the user.create
event.


test/user_works/handlers_test.clj
```diff
@@ -37,11 +37,44 @@
...
-     (testing "adds a message to the create events channel"
-       (async/alt!!
-         (async/timeout 1000) (is false "No create event received after 1 second.")
-         creates-chan ([event] (is (= (get-in resp [:user :id])
-                                      (:id event))))))))
+      (testing "adds a message to the create events channel"
+        (async/alt!!
+          (async/timeout 1000) (is false "No create event received after 1 second.")
+          creates-chan ([event]
+                        (is (= (get-in resp [:user :id])
```
cap10morgan: Is this a redundant check now?

test/user_works/handlers_test.clj
```diff
@@ -37,11 +37,44 @@
...
+                                :zip "70119"
+                                :ocd-divisions
+                                #{"ocd-division/country:us/state:la"}}
+                               :mailing
+                               {:street "2659 Iberville St."
+                                :city "New Orleans"
+                                :state "LA"
+                                :zip "70119"
+                                :ocd-divisions
+                                #{"ocd-division/country:us/state:la"}}}}
```
cap10morgan: Should we use a different address here to make sure the system is storing them separately?

ericnormand: Good idea.

cap10morgan:
Assuming test run is green, LGTM.

PR #22, opened 2016-02-24
[SD WM] Move the OCD-ID fallback into user-works

sduckett:
The second half of [turbovote-web pull request](https://github.com/democracyworks/turbovote-web/pull/91) and should be deployed at the same time.

- Blocked by [a bifrost pull request](https://github.com/democracyworks/bifrost/pull/3) and a [user-http-api pull request](https://github.com/democracyworks/user-http-api/pull/12).
- [Pivotal Card](https://www.pivotaltracker.com/story/show/114324515)


src/user_works/handlers.clj
```diff
@@ -43,9 +52,9 @@
       (if-not (= :timeout response)
         (if-let [ocd-divisions (:ocd-divisions response)]
```
7maples: Are we confident that we never get an empty set out of the response?

src/user_works/handlers.clj
```diff
@@ -33,6 +33,15 @@
                  :server (str "Unknown server error: "
                               (.getMessage runtime-exception))}}))))
 
+(defn fallback-ocd-id [state]
+  (let [country-ocd-id "ocd-division/country:us"
```
7maples: Doc string s'il vous plait

src/user_works/handlers.clj
cap10morgan: Good point. I don't think I've seen an empty set come back, but it certainly could. We should fall back to country and state then too!

src/user_works/handlers.clj
sduckett: certainly.

src/user_works/handlers.clj
sduckett: I'll have a test.

sduckett:
I've addressed your comments, @7maples.

7maples:
LGTM

PR #23, opened 2016-04-14
Add language preference

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/114353005)

@sduckett would you care to take a look?


resources/schemas/005-add-language-preference-and-migration.edn
```diff
@@ -0,0 +1,29 @@
...
+ {:db/id #db/id[:db.part/db]
+  :db/ident :user/set-default-language-preference
+  :db/fn
+  #db/fn {:lang "clojure"
+          :params [db]
+          :code
+          (let [without-pref
+                (datomic.api/q
+                 '{:find [?e]
+                   :in [$]
```
sduckett: I'm a little uncertain on how this function works. Can you help me understand how $ is bound when this function is invoked?

tank157: I think either by convention or by requirement, $ is always bound to the current database value. When writing your own query functions, you can leave off the `:in [$]` if there are no other parameters (it's assumed), but if, for instance, I wanted to find all the people with a certain value and wanted to pass that in, it would look like `:in [$ ?my_value]`.

sduckett: OK, that makes more sense. I think I've only seen it in the `:in [$ ?my_value]` form.

sduckett:
LGTM

resources/schemas/005-add-language-preference-and-migration.edn
tank157: Yeah, I needed the $ for the `(missing?...)` call. You can read more about it here: http://docs.datomic.com/query.html (just search on missing, there's no anchor)

PR #24, opened 2016-05-11
Replace either w/ conditional in schema

cap10morgan:
`either` was deprecated in the schema library and the better error messages alone make `conditional` a
huge win over it anyway.

I used `conditional` to help debug a problem I was seeing in krakenstein and it was very helpful. So I figured I put it up in a PR for others to benefit from too.


sduckett:
@cap10morgan Do you have an example of how are the error messages better?

cap10morgan:
Not handy, but basically the difference is `either` tells you *that* it didn't match but `conditional` tells you *why* (like most validation errors do in schema).

sduckett:
LGTM

sduckett:
Let's merge this one, @cap10morgan 👍

PR #25, opened 2016-06-17
Stop using status attribute and retract deleted users instead

cap10morgan:
[Pivotal card](https://www.pivotaltracker.com/story/show/120239959)

~~Card is un-estimated~~ but this was a Friday project. I retro-estimated it at 2 points based on the amount of work it required.

This is more in line with what we do with Datomic entities elsewhere, better fits with users' expectations when they request deletion, and was signed off on by product (they agreed that we should retract the Datomic entities of users that request deletion). It also simplifies the code and data model.

I tested this locally in krakenstein and everything still seemed to work.

Once deployed, I'll retract any users in the production database whose status is anything other than `:active`. I'll also retract the active users' `status` attributes.


marlabrizel:
LGTM

PR #26, opened 2016-07-07
adds simple user data export to CSV (in REPL) functionality

ddellacosta:
This comes out of research card: https://www.pivotaltracker.com/story/show/116584169

This includes some basic utils to export user data, including a function that simply dumps all user data available in user-works (all users with *any* hostname, as I understand it) to a CSV file.


danielglauser:
LGTM

dev-src/user_works/dev.clj
```diff
@@ -0,0 +1,138 @@
...
+    [clj-time.core :as t]
+    [clj-time.format :as tf]
+    [clojure.java.io :as io]
+    [clojure.data.csv :as csv]
+    [datomic.api :as d]
+    [turbovote.resource-config :refer [config]]))
+
+(defn find-users-with-email
+  [db email]
+  (->> (d/q '{:find [?user]
```
cap10morgan: `:find [[?user ...]]` would eliminate the need for the mapping over `first` on line 17. It stops wrapping every result in another single-element collection.

dev-src/user_works/dev.clj
```diff
@@ -0,0 +1,138 @@
...
+              :in [$ ?hostname]
+              :where [[?user :user/hostname ?hostname]]}
+            db hostname)
+       (into [] (format-user-record db))))
+
+(defn all-users'
+  [db]
+  (d/q '{:find [?user]
+         :in [$]
+         :where [[?user :user/hostname]]}
```
cap10morgan: I would use `:user/id` over `:user/hostname` here. All users are guaranteed to have id's, but perhaps not hostnames.

dev-src/user_works/dev.clj
```diff
@@ -0,0 +1,138 @@
+(ns user-works.dev
```
cap10morgan: I would prefer something more focused like `user-works.export` here.

cap10morgan:
If we add a simple `-main` fn this can be handy to use outside the REPL too. (e.g. `lein run -m user-works.export all-users.csv`)

ddellacosta:
@cap10morgan Finally got this tested and pushed.  Check out when you have a chance please!

dev-src/user_works/export.clj
```diff
@@ -79,16 +79,33 @@
...
+   (case filename
+     "--help" (println
+"Pass in a filename like so:
+  $ lein run -m user-works.export all-users.csv
+
+Without a filename, this will generate a filename in the format
+'users-YYYY-MM-DD.csv'
+
+Please note that this requires a datomic uri to be set
+for the user-works partition in resources/config.edn
```
cap10morgan: Setting the prod Datomic URI in `resources/config.edn` in dev is probably going to get us in trouble at some point. Could we get it from a `DATOMIC_URI` env var instead? That's what electorate-works' verify code does.

cap10morgan:
One last concern about putting the prod Datomic URI into dev config.edn. Other than that, LGTM!

PR #27, opened 2016-07-20
make room for referral codes

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/119561725)

This is the backend work so we can store referral codes with users.

Assigning to @7maples


7maples:
LGTM

PR #28, opened 2016-07-22
User Export for Turbovote Admin Dashboard, v1

ddellacosta:
https://www.pivotaltracker.com/story/show/119558509


project.clj
```diff
@@ -20,7 +20,12 @@
                  [org.immutant/core "2.1.5"]
                  [democracyworks/kehaar "0.5.0"
                   :exclusions [commons-codec]]
-                 [prismatic/schema "1.1.2"]]
+                 [prismatic/schema "1.1.2"]
```
tank157: I suppose since i got stung with this too, so do you. We can upgrade prismatic/schema to "1.1.3" which just came out 3 days ago.

src/user_works/entities/user.clj
```diff
@@ -45,6 +45,20 @@
...
        db ocd-id))
 
+(defn find-by-hostnames
+  [db hostnames]
+  (d/q '{:find [[?user ...]]
+         :in [$ [?hostnames ...]]
+         :where [[?user :user/hostname ?hostnames]]}
+       db hostnames))
+
+(defn all-users
```
tank157: Can you talk about where the need for this fn is coming from? The card doesn't call for a total user export, just partner specific. Is this something we want to allow?

src/user_works/entities/user.clj
```diff
@@ -45,6 +45,20 @@
...
+  [db]
+  (d/q '{:find [[?user ...]]
+         :in [$]
```
tank157: Fun fact, if you don't have any parameters to the query, you can elide the entire `:in [$]` line.

test/user_works/test_helpers.clj
```diff
@@ -0,0 +1,11 @@
+(ns user-works.test-helpers
+  (:require [datomic-toolbox.core :as db]
+            [datomic.api :as d]))
+
+(defn test-db-fixture
```
tank157: We mostly call this `with-fresh-db` in other projects. I don't know that we're consistent everywhere for sure, but maybe we should stick to the same name so we can grep it later for paying down the technical debt? Is there another project that uses this name?

src/user_works/handlers/export.clj
```diff
@@ -0,0 +1,82 @@
...
+            [clojure.java.io :as io]
+            [clojure.tools.logging :as log]
+            [datomic-toolbox.core :as dt]
+            [datomic.api :as d]
+            [me.raynes.fs :as fs]
+            [turbovote.resource-config :refer [config]]
+            [user-works.entities.user :as users]))
+
+(defn format-date-of-birth
+  [dt]
```
tank157: This would be a great spot for a docstring showing what the output of this formatter looks like.

src/user_works/handlers/export.clj
```diff
@@ -0,0 +1,82 @@
...
+  (if dt
+    (tf/unparse (tf/formatter :date) (tc/from-date dt))
+    dt))
+
+(defn format-user-record
```
tank157: This function starts with the most raw form of the entity (the kind you call d/entity on, so just the `{:db/id 233454645}` form. It then does a bunch of formatting to the retrieved entity form, but if I'm not mistaken, that's the same transformation we do in `entity->user`. So it's kind of duplicative. If there's a reason for this, I'd like to know, but in general we try to conform to things outside of the `entities` namespaces dealing with the simplified representations.

So what I'm suggesting is that whatever is querying the database (lazily) applies the `entity->user` transformation, and then this formatting code starts with something where addresses have already been flattened, and applies updates to that. It also means the `ordered-keys` and whatnot would have simplified names, `:id` instead of `:user/id`, etc.

That also highlights a bug in this code, `:user/address` isn't a single address, there might be as many as 3 addresses in there, and this code doesn't seem to take that into account from what I can tell. If we're only exporting the physical address, that's under the key `:registered` after `entity->user` has been run, or is the one with `:address/type` == `:registered`.

src/user_works/handlers/export.clj
```diff
@@ -0,0 +1,82 @@
...
+                     (users/find-by-hostnames (dt/db) hostnames)
+                     (users/all-users (dt/db)))
+                   (into [] (format-user-record (dt/db))))
+        csv-tmpfile (fs/temp-file "user-export-")
+        csv-stream (generate-csv users csv-tmpfile)
+        bucket (config [:aws :s3 :bucket])
+        credentials (config [:aws :creds])
+        key (gen-filename (clojure.string/join "-" hostnames))]
+    (s3/put-object credentials bucket key csv-tmpfile)
+    (try
```
tank157: I'd like to see this try block wrapping the s3/put-object call at least, but also probably the generate-csv call too. Both of those seem like spots that can/will error out at some point in some way, and we don't want to leave orphaned temp files around.

test/user_works/handlers/export_test.clj
```diff
@@ -0,0 +1,53 @@
...
+  (let [bytes (ByteArrayOutputStream.)
+        csv (export-handlers/generate-csv users (io/output-stream bytes))]
+    (io/reader (io/input-stream (.toByteArray bytes)))))
+
+(deftest exports-csv
+  (let [user1 {:first-name "Buffy"
+               :last-name "Summers"
+               :hostname "test.turbovote.org"
+               :language-preference :es
+               :addresses {:headquarters {:street "123 Main St"
```
tank157: With my earlier concerns about the code not properly handling addresses, this is a good example. In theory we can store addresses under just about any key, but in practice `:headquarters` isn't one of them. And while this in theory "works", unless `:headquarters` is ALSO where Buffy is registered to vote, probably not what we want. Some of this needs clarification in the card around addresses, namely, do we only want a single address in the export, and if so, is that their physical/registered address? Or do we want to also have space for mailing addresses in the export? But I'm pretty sure if we want a single address, we'll want to make sure which one it is.

tank157:
Ok, I think I'm done with my initial comments pass.

src/user_works/entities/user.clj
ddellacosta: Oh, good question. So I double-checked this with Katy just now after I read your question, and it's a go from her perspective, but let me give you some background:

- the original version was basically a dump of all users that I passed directly to Katy. This is mostly just a carry-over from there.
- the idea is that only folks who are members of an "admin" group can get access to this. Everyone else has to be a member of one or more partner-specific groups or else they get nothing, effectively.

Basically, it was relatively trivial to carry this functionality over, and so we did, but I didn't consider whether or not it was actually desired until I read your question--so I asked Katy whether or not we actually want this functionality or if we should remove it, and she said it seems quite useful.  So, assuming you don't have any other issues with it, I think it should be included.

src/user_works/entities/user.clj
ddellacosta: I love deleting code! Thanks for the info.

test/user_works/test_helpers.clj
ddellacosta: I think I just made this up--sounds good to me!

src/user_works/entities/user.clj
tank157: Ok, cool, just wanted to make sure we had cause to include it.

src/user_works/handlers/export.clj
ddellacosta: Yeah, I will never object to adding a doc-string, good call. I also changed the name to reflect better what the function actually does.

src/user_works/handlers/export.clj
ddellacosta: Okay, so I've mostly got this fixed and matching what you've suggested, but I've diverged a little for what I think are good reasons:

* I'm making sure we are returning entities from the query functions. I think it wasn't clear to me what the convention was, as for example (`find-by-ocd-id`)[https://github.com/democracyworks/user-works/blob/aaf1a08aa031ee4787586c73e03cb84e27662300/src/user_works/entities/user.clj#L39-L46] returns IDs only, whereas the function above it returns the entity.

- The bug regarding simply returning the first address is fixed; we specifically pull out `:mailing` and `:registered` type addresses and put them into the root user map for formatting more easily later on.

The one thing I have diverged on is where I flatten and format the addresses: after reviewing the code, there didn't seem to be enough useful overlap in terms of the functionality in `entity->user` or anywhere else in the current entities namespaces to do what I needed to do for the purposes of this CSV export; for the purposes of the export, it doesn't seem to matter so much whether or not the keys are de-namespaced, and the address munging still involves mostly the same amount of work regardless of whether or not `entity->user` is called first. Perhaps it still makes sense to move some of this code into one of the entities namespaces, but it wasn't clear to me what would make sense there, since the map-restructuring I'm doing is very much specific to this export.

In any case, I'm happy to consider this first of course--let me know if you still I'm off-base here.

src/user_works/handlers/export.clj
ddellacosta: I think what I've done should address this--I've specifically wrapped the chunk of AWS S3-specific code to catch issues there, and otherwise wrapped the overall export function in a larger try/catch block to catch anything else that may happen outside of that.  Let me know if you think there is a better way to structure this to catch the sort of errors you are thinking may arise.

ddellacosta:
@tank157 I think most of your concerns are addressed now Troy--let me know if you still have any points of contention with https://github.com/democracyworks/user-works/pull/28#discussion_r72506536, but otherwise I believe we should be good to go.

FYI I did go over the address stuff with Katy to make sure we were doing what was expected in the user export, and Katy also caught one thing I missed: I had failed to make sure we only exported the year for the date-of-birth--that should be good now.

dev-src/user_works/export.clj
```diff
@@ -1,157 +1,49 @@
...
-;; mailing-subscription
-
-;; Not sure?--what's the difference with the address taken from user-works
-
-;; mailing-street
-;; mailing-street-2
-;; mailing-city
-;; mailing-state
-;; mailing-zip
+     (let [conn (connect-to-user-works-db)]
```
tank157: I'm curious about the reasoning behind this make/release a connection code, it's rather unlike what we do elsewhere which is just re-use the same connection you get from (datomic-toolbox.core/connection), and that one is already configured, and is configured by turbovote-resource.

dev-src/user_works/export.clj
```diff
@@ -1,157 +1,49 @@
...
+
+(defn export-users!
+  [conn hostnames filename]
+  (let [db (d/db conn)
+        users (case hostnames
+                ":all" (users/all-users db)
+                (users/find-by-hostnames db (str/split hostnames #"\s")))]
+     (-> (into [] (export-handler/format-user-record) users)
+         (export-handler/generate-csv! filename))
+    (println "Exported users for " hostnames " in file " filename)))
```
tank157: Can this be changed into log/debug please?

ddellacosta: Oh sure, except it's mean to be dumped out on the command-line--does it matter (I honestly don't know)?

dev-src/user_works/export.clj
ddellacosta: Oh, I was just seeing this hanging when running it from the command-line and thought maybe hanging onto the connection was why.  Didn't work though, so happy to revert it if you think that's appropriate.

In general I was also trying to be more explicit about the connection/db here as well.

ddellacosta:
@tank157 Hey Troy, I got rid of the `release` call and switched it to `System/exit` which was apparently what I was lacking.  I figured this out by reviewing some other dev-src code.  Otherwise, it seems like we use `println` in general when stuff is meant to be called on the command-line--is this okay or should we be moving to `log/debug` across the board?  Also, I didn't see a lot of consistency with how we load up datomic in various `dev-src` namespaces, but again will change if there is a best practice there I'm missing--just didn't seem like we needed or wanted what `datomic-toolbox` provides in this case.  Also see https://github.com/democracyworks/user-works/pull/26 for more context re: the first iteration of this.

dev-src/user_works/export.clj
tank157: There's probably a lot to revisit someday about datomic-toolbox, and configuration, but for now probably makes more sense to keep with what we're doing to make future transitions easier.

There's also some hard won lessons about datomic buried in that code, namely that reconnecting to the db can sometimes be an expensive operation, despite what the docs and Cognitect tell us. We were seeing somewhat frequent 5 second pauses when we were reconnecting all the time. So the current way it's done in datomic-toolbox helps alleviate that, but this would subtly introduce it back in this one area.

dev-src/user_works/export.clj
tank157: Well, unless it's in a `dev-src` package, pretty much every output statement is ultimately going to get run in a web app in wildfly. To manage that output, we're using logging. So this either needs to go all together, or get turned into logging.

dev-src/user_works/export.clj
ddellacosta: Oh gotcha.  Still, this is all good stuff for me to keep in mind, useful information that I wasn't necessarily aware of--thanks.

test/user_works/handlers/export_test.clj
ddellacosta: Troy sorry I think I missed responding to this yesterday because I changed this code and github hid the comment.

In any case, I think I've addressed exactly your concerns--we now are selection the `:registered` and `:mailing` addresses explicitly, and assume a single address for each.  So I think we should be good now.  Good catches all around--I don't have the domain knowledge yet to see these things, so thanks!

dev-src/user_works/export.clj
```diff
@@ -1,157 +1,49 @@
...
 
 (defn -main
   ([]
-   (let [filename (gen-filename)]
-     (-main filename)))
-  ([filename]
-   (case filename
+   (-main ":all"))
+  ([hostnames]
```
tank157: What about passing the datomic db URI in as a parameter and not as an env var? It's comparable to how we pass the rabbit config in as a parameter when we do the election and authority imports. Just a thought to simplify how this would get called, all parameters, instead of a mix of parameters and env vars.

src/user_works/handlers/export.clj
```diff
@@ -0,0 +1,156 @@
...
+   :user/language-preference
+   :user/hostname])
+
+(defn generate-csv!
+  "Generates a CSV file based on formatting the user values in users
+  according to the key ordering in ordered-keys. Takes in a thing
+  which is coercable to a writer (which in terms of current use-cases
+  includes files or filename strings) and returns this value."
+  [users coercable-to-writer]
+  (let [formatted-users (->> users
```
tank157: I'm still not really thrilled with all this exposure of the entity type outside of the entities namespace. It still feels redundant to a lot of what the `entity->user` transformation does, and breaks from the intended use patterns. Not that that would automatically be an issue, because we do break in other places, but only really when we need knowledge of the entity as a thing with a `:db/id` for things like updates.

But, `user-works` already doesn't follow a lot of the later learned idioms, so I can let it slide for now, and it'll probably all get revisited sometime post election as a technical debt paydown.

dev-src/user_works/export.clj
ddellacosta: Sure, happy to change it--just to be clear, I was following Wes's suggestion [here](https://github.com/democracyworks/user-works/pull/26#discussion_r70682351) in the first place.

src/user_works/handlers/export.clj
```diff
@@ -0,0 +1,156 @@
...
+  (try
+    (let [users (->> (if-not ((set hostnames) "admin")
+                       (users/find-by-hostnames (dt/db) hostnames)
+                       (users/all-users (dt/db)))
+                     (into [] (format-user-record)))
+          csv-tmpfile (fs/temp-file "user-export-")
+          _ (generate-csv! users csv-tmpfile)
+          s3-filename (gen-filename (clojure.string/join "-" hostnames))
+          export-url (upload-to-s3! s3-filename csv-tmpfile)]
+      (try
```
tank157: This isn't really what I was trying to convey here. What I'm concerned is that from the point `csv-tmpfile` is bound, the next several operations all have the possibility of throwing exceptions and `csv-tmpfile` won't be deleted. This isn't really changing that.

As ugly as it would be, I suggest you stop the let bindings immediately after `csv-tmpfile` is bound. Then inside the let body, start a `(try ... (finally (.delete csv-tmpfile)))`. If you need a second let form inside, ok, but I bet the rest of the original let bindings could be reworked into a threading.

dev-src/user_works/export.clj
tank157: Hah, fun, I guess we're all over the place on this. `dev-src` does tend to be more like the wild west. Ok, consider my suggestion totally optional.

ddellacosta: Okay, fixed! And I think this is cleaner too--prefer to pass this stuff in vs. having a mix of params and env vars as you say, and also just like that it's an explicit arg vs. hidden tucked away inside a function somewhere.

src/user_works/handlers/export.clj
ddellacosta: Got it--I don't think I really understood your point re: namespaced keywords the first time around, but I've changed this so it is pre-processed by `entity->user` first (and this actually simplified the code a bit, so furthermore I think I must have been mistaken in my initial assessment regardless).  Thanks for your patience and for taking the time to explain this to me in depth!

ddellacosta:
@tank157 Okay Troy, think we may be good here--I believe I've addressed everything you brought up at this point. Let me know!

tank157:
LGTM

dev-src/user_works/export.clj
cap10morgan: Yeah, the deal here is that w/o going to a full on CLI arg parsing library, they just don't scale that well. So if there are 2 or fewer CLI args, then fine. Beyond that, I start reaching for env vars.

PR #29, opened 2016-07-28
move aws cred location to top level in consul's kv store

ddellacosta:


ddellacosta:
Nevermind--per Slack we decided to have `user-works` have its own creds, so it is good as it is.

PR #30, opened 2016-08-01
Add in referral code to export

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/119561947)

Assigning to @ddellacosta


test/user_works/handlers/export_test.clj
```diff
@@ -57,4 +58,6 @@
             (is (= (get-in user2 [:addresses :registered :zip])
                    (nth csv-user2 10)))
             (is (= (get-in user2 [:addresses :mailing :state])
-                   (nth csv-user2 14)))))))))
+                   (nth csv-user2 14)))
+            (is (= "foo" (:referral-code user1)))
+            (is (nil? (:referral-code user2)))))))))
```
ddellacosta: I think it would be useful to confirm that the output of the `csv-user1`/`csv-user2` vars include the referral-code here--as it is this simply tests that we have the data in the user maps we've created, not whether or not that made it into the CSV output.

tank157: Good point, I think I intended to test `csv-user-1/csv-user2` and miswrote it. I'll fix that, and if tests still pass, consider it good.

ddellacosta: Sounds good to me!

ddellacosta:
So yeah, leaving the test thing aside LGTM!

PR #31, opened 2016-08-02
broadcast full user

tank157:
Instead of just broadcasting the user id, we now broadcast the full user. This allows other services a chance to do something with the user data, and so that the broadcast of the full electorate doesn't try to look up a now deleted user.

Not as many use cases for this as there are for election, but wanted to maintain the same broadcast standards between both.

Assigning to @ddellacosta


ddellacosta:
LGTM

PR #32, opened 2016-08-09
Add formatted timestamps to export

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/127932173)

Assigning to @7maples


7maples:
I would love to generate an export to test this, but cannot find any documentation in the readme or torchpad. Do you have some instructions I could try, and then create documentation out of?

tank157:
I do not. I admittedly have never done that myself, despite adding a few
fields to the exports. You'll need to check with Daniel or Dave to get
those instructions.

tank157:
But what i have done is added a println of the pr-str'd users to the export
test and looked at it that way in the REPL output.

7maples:
LGTM

tank157:
The changes to export will likely clobber this work as it exists, so since it's pretty small, will re-implement over there.

PR #33, opened 2016-08-10
adds user-ids-by-hostnames handler

ddellacosta:
https://www.pivotaltracker.com/story/show/128103513

Also tested manually using docker/rabbitmq


ddellacosta:
So, I just realized something, which is that I'm not returning these in a vector, which is a requirement of the card.  It's trivial to do this, but does it really matter?  The only thing that cares about the ordering should be the user of the return value of this endpoint, and it can order it however it wants.

cc @tank157 @danielglauser

src/user_works/queue.clj
```diff
@@ -48,14 +48,21 @@
...
                                                channels/export-requests
                                                channels/export-responses)
 
+          by-hostnames-ch (wire-up/incoming-service
+                        connection
+                        "user-works.user.user-ids-by-hostnames"
+                        (config [:rabbitmq :queues "user-works.user.user-ids-by-hostnames"])
+                        channels/by-hostnames-requests
+                        channels/by-hostnames-responses)
+
```
cap10morgan: Can you fix the indentation here?

cap10morgan:
Small indentation fix needed. After that, LGTM.

tank157:
@ddellacosta I concur, it's not so important that this component return an ordered sequence, the handler in `turbovote-admin-works` that uses these ID's can order them as it sees fit.

PR #34, opened 2016-08-11
add validation to user-ids-by-hostnames handler

ddellacosta:


ddellacosta:
@cap10morgan Sorry I should have added this to the last one--would you mind taking a look?

src/user_works/handlers.clj
```diff
@@ -213,8 +213,12 @@
      :error {:type :not-found}
      :id id}))
 
+(def user-ids-by-hostnames-schema
+  {:hostnames (s/cond-pre (s/enum :all) [s/Str])})
```
cap10morgan: Sorry I didn't catch this before. Yes, schemas are good.

Also: should hostnames be a vector or a set? Does order of the hostnames in the request matter?

src/user_works/handlers.clj
```diff
@@ -213,8 +213,12 @@
...
+
 (defn user-ids-by-hostnames
   [{:keys [hostnames] :as message}]
```
cap10morgan: Can we spit out a debug-level log message along the lines of `"Received user-ids-by-hostnames request:" (pr-str message)` here?

test/user_works/handlers_test.clj
```diff
@@ -835,4 +835,14 @@
...
+
+    (testing "validates incoming message"
+      (is (thrown? java.lang.RuntimeException
+                   (handlers/user-ids-by-hostnames {:hostnames "foo"})))
+      (is (thrown? java.lang.RuntimeException
+                   (handlers/user-ids-by-hostnames {:hostnames [:foo]})))
+      (is (thrown? java.lang.RuntimeException
+                   (handlers/user-ids-by-hostnames {:hostnames :not-all})))
+      (is (thrown? java.lang.RuntimeException
+                   (handlers/user-ids-by-hostnames {}))))))
```
cap10morgan: Should we use `thrown-with-msg?` here and verify that it's a schema validation exception?

src/user_works/handlers.clj
ddellacosta: Order doesn't matter, can and probably should be a set.

ddellacosta:
@cap10morgan Take another look?

cap10morgan:
LGTM

PR #35, opened 2016-08-12
Stream users

tank157:
[Pivotal Card](https://www.pivotaltracker.com/story/show/128103827)

Assigning to @ddellacosta


ddellacosta:
LGTM. Ship it!!

danielglauser:
You'll need a merge commit but what I see LGTM.

src/user_works/handlers.clj
```diff
@@ -213,13 +213,18 @@
...
 (defn user-ids-by-hostnames
   [{:keys [hostnames] :as message}]
+  (log/debug "Handling request for user-ids for:" (pr-str message))
+  (s/validate user-ids-by-hostnames-schema message)
   {:status :ok
    :user-ids
-   (case hostnames
-     :all (user/all-uids (db/db))
-     (user/find-uids-by-hostnames (db/db) hostnames))})
+   (if (seq hostnames)
```
ddellacosta: So, I would push back against this change.  There are security implications for what we expose if we get a non-seq, and we don't want to accidentally return all the values on nil when we should only be doing this for `:all`.  So this needs to at least be explicitly checking that we are getting `:all` vs. something else when we return all the users.  As it was before, it would do the right thing and return nothing when an empty hostnames seq was passed to `find-uids-by-hostnames`, which I think is preferable.

danielglauser: Good catch @ddellacosta!

src/user_works/handlers.clj
```diff
@@ -213,10 +213,32 @@
      :error {:type :not-found}
      :id id}))
 
+(def user-ids-by-hostnames-schema
+  {(s/optional-key :hostnames) #{s/Str}})
```
ddellacosta: [Per comment below](https://github.com/democracyworks/user-works/pull/35/files#r74815184), this should not be optional, and should either be `:all` or a set of strings.

src/user_works/handlers.clj
tank157: I'm talking myself back into `:all` here. I'm not overly fond of it. We're going to have to do authorization on the hostnames anyway at some point, I don't know that `:all` is really any better. But, with `:hostnames` being optional, I still have to pass an empty map, and that's ugly too. So I'll put `:all` back in for now, but we may want to consider something else for the future.

ddellacosta: Well what I'm most concerned with here is security--we need to be explicit *somehow*, and maybe that involves something other than using a keyword like `:all` (a different endpoint somewhere else?  I dunno) but I'm not sure how else to handle this at this point.

tank157: I'm just not real fond of having `:hostnames` point to potentially two different things. It just makes all code dealing with it uglier. But perhaps we can handle that with a schema coercion later.

PR #42, opened 2016-08-19
add paged export endpoint

tank157:
Adds a paged export endpoint that returns a coll of users in a single message, based on the input user-ids.


ddellacosta:
LGTM!

PR #45, opened 2016-09-09
Add uuids to error

7maples:
[Pivotal Card](https://www.pivotaltracker.com/story/show/129862097)

Adding a user's uuid to the geocode timeout error was more difficult than anticipated, since at the time of the error we have not yet created the user, therefor we have no id. So we had to generate an id sooner and then use that when creating the user in Datomic. Updated a bunch of tests as well to accommodate this change.


src/user_works/handlers.clj
```diff
@@ -52,7 +52,7 @@
                  (clojure.string/lower-case state))))))
 
 (defn geocode-address
-  [address]
+  [id address]
```
cap10morgan: Since this is a user id and not an address id, can well call this `user-id` in here?

test/user_works/entities/user_test.clj
```diff
@@ -35,15 +38,17 @@
       (is (= tx (users/->datomic-tx-datom user)))))
 
   (testing "sets default language-preference if not provided"
-    (let [user {:first-name "Buffy"
+    (let [id (users/generate-user-id)
+          user {:id id                :first-name "Buffy"
```
cap10morgan: Let's put `:first-name "Buffy"` on the next line.

test/user_works/handlers_test.clj
```diff
@@ -434,10 +434,11 @@
                 :dob #inst "1981-07-18"
                 :us-citizen true}
           user-id (get-in (create-user (async/chan) user) [:user :id])
-          resp (read-user {:id user-id})]
+          resp (read-user {:id user-id})
+          user-keys-without-id (-> (remove #(= :id %) user/user-keys) vec)]
```
cap10morgan: Since this is just threading one value through one fn, can we make it either:

```clojure
(->> user/user-keys
     (remove #(= :id %))
     vec)
```

OR

```clojure
(vec (remove #(= :id %) user/user-keys))
```

src/user_works/handlers.clj
```diff
@@ -64,7 +64,7 @@
         (if (seq ocd-divisions)
           (merge address {:ocd-divisions ocd-divisions})
           (merge address {:ocd-divisions fallback-divisions}))
-        (do (log/error "Response timeout calling address-works to geocode address:" address "for user:" id)
+        (do (log/error "Response timeout calling address-works to geocode address:" address "for user:" user-id)
```
cap10morgan: Actually, now that I take another look at this, I realize that `geocode-address` doesn't need the `user-id` expect for inserting it into this log statement (and then only on an error).

I think a better approach would be to:

1. Have `geocode-address` just take the `address` arg since that's really all it needs to do its job.
1. Have it return a value indicating a timeout to its caller.
1. Have the caller (which presumably has the user's id) log an error like the above.

danielglauser:
Moved work [here](https://github.com/democracyworks/user-works/pull/55).

PR #47, opened 2016-09-15
Phone Number Search

ddellacosta:
[Tracker Card](https://www.pivotaltracker.com/story/show/126043967)

@tank157 Not really ready to PR but pinging you for discussion.

A few questions that arose when working on this, thinking about what we had talked about previously.

- should I refactor this to allow for a batch of email addresses or phone numbers (or district-divisions for that matter) to be searched on?
- do we want this to be additive for phone numbers + email addresses (and district-divisions for that matter?)

I guess I'll start with that...let's talk!


tank157:
In regards to the question, I think we're fine with non-batching capabilities right now, and with non-combinative searches too. emails and phone numbers tend to be pretty identifying on their own. At some point, we may want something like name + location searches, or other combinations, but I think it's fine to wait to write that when we get an actual use case that calls for it.

src/user_works/entities/user.clj
```diff
@@ -134,37 +135,65 @@
     {:db-after (:db-after @(db/transact [user-tx]))
      :user-id (:user/id user-tx)}))
 
+(def search-users-base
```
tank157: I'm not sure I find this useful to abstract out, it just makes the queries below a bit more difficult to reason about.

src/user_works/entities/user.clj
```diff
@@ -134,37 +135,65 @@
...
+  [db query query-term hostnames]
+  (-> query
+      (update :in conj '[?hostnames ...])
+      (update :where #(apply conj '[[?user :user/hostname ?hostnames]] %))
+      (d/q db query-term hostnames)))
+
+(defmulti search-query
+  (fn [_ search-terms] (into #{} (keys search-terms))))
+
+(defmethod search-query #{:search-email :hostnames}
```
tank157: Now this I do like, using the multi-method here is nice.

src/user_works/entities/user.clj
ddellacosta: Yeah...I think I agree with you.  This part in particular was experimental and I'm not sure it gives us a lot of value in this case.  If we did want this to be compositional (`and` vs. `or` query built up programmatically) then I suspect we could get more out of it, but as it is I think I'm on the same page--going to inline this stuff and refactor a bit.

src/user_works/entities/user.clj
ddellacosta: Cool--especially if we want to keep addition conditions here this aspect of it could be useful. If we decide to change the query significantly could get in our way, but I guess we'll figure that out later on.

PR #53, opened 2016-10-21
Remove deprecated export code

danielglauser:


src/user_works/queue.clj
```diff
@@ -49,13 +48,6 @@
                                           channels/delete-requests
                                           channels/delete-responses)
 
-          ;; DEPRECATED to be replaced by streaming response to other service per
-          ;; https://www.pivotaltracker.com/story/show/128103827
-          exports-ch (wire-up/incoming-service connection
-                                               "user-works.user.export"
-                                               (config [:rabbitmq :queues "user-works.user.export"])
-                                               channels/export-requests
```
tank157: Oh, also need to remove the channels from channels.clj

PR #54, opened 2016-11-29
Use kehaar `configure!` to set up queues and handlers

tie-rack:
kehaar's coming `configure!` function uses a map to set up queues and channels and functions writing a bunch of `wire-up` code. Here's what it would look like in user-works.

The only core.async channels that need to be created anymore are the outgoing event channels. The rest of them are created and managed by kehaar.

`augment-address` has changed from being a function created with `wire-up/async->fn` to merely being `declare`d because it's now set up by `configure!`. This necessitates using `with-redefs` in the tests (instead of the weird channel routing that was happening before), and though I am usually anti-`with-redefs`, this is a situation I'm okay with: when a function's purpose is to return results from a outside service, and when it actually makes the tests read more clearly.

Since configuration of handlers deals with symbols that reference functions, many of the handlers needed just a little change: being wrapped in `handle-errors` and partialized with the event channel it uses.

Looking at [the diff that ignores whitespace](https://github.com/democracyworks/user-works/pull/54/files?w=1) will give a slightly better view of the test changes.

Does this look better? I think so.


test/user_works/handlers_test.clj
```diff
@@ -628,123 +623,112 @@
...
+                                             :state "CO"
+                                             :zip "80222"}}}]
+        (with-redefs [user-works.handlers/address-works-timeout 500]
+          (let [user-resp (create-user* (async/chan) user)]
+            ;; Logging this because it failed several times in a row (in CI
+            ;; builds only) with {:status :error}; so if that happens again
+            ;; I want to catch it.
+
+            ;; Who wrote the comment above? Did you ever catch it? - WSM
+            (log/info "user-resp:" (pr-str user-resp))
```
cap10morgan: Turns out I wrote both comments, with enough time separation to forget (but git doesn't forget). So I think, if you don't feel that it's too out-of-scope, we could drop the comments and the `log/info` from here. I haven't seen this issue in quite some time.

ddellacosta:
(For the most part, by "service," in this context, I mean something mapping onto a single queue for use in a given micro-service.)

Stuff I dislike:

1) `configure!` is opaque and implicitly manages life-cycle--it's not simply configuring things, it's connecting to rabbitmq, setting up queues and channels for all services for a given micro-service, etc., implicitly.  `configure!` is the wrong name, but I also don't think this should be handling the rabbitmq connection implicitly regardless.

2) The way external service access functions (`augment-address`) are set up remains a namespace-globally scoped var with implicit lifecycle and is pretty magical.  I think the rationale for why this is justified and why `with-redefs` is acceptable here is weak, and ideally this would simply be a channel passed in to the functions that need it, somehow.

(This seems like the toughest nut to crack but I still don't think this is ideal.  That said I'm really not sure how to do this The Right Way either.)

3) I would rather it still be possible to configure specific services independently for testing and one-off uses vs. requiring a big ole' blob of a configuration map any time we want a single kehaar service (or will there another clean way to do this?).  In general I think it will serve us better to have a bunch of small modular configuration functions for setting up services vs. a big one...again, I believe using component or something like it to manage life-cycle AND configuration of individual services is the right approach, because as is this awkwardly couples configuration and initialization and leaves shutdown as something for the client user to figure out.

Stuff I like:

1) Removal of the need to declare all the other channels, and passing those in to the functions that use them.

2) My beefs aside, it's nice that we no longer have the kinda weird `async->fn` stuff for external services, and despite my complaints about magic I'm not sure how we'd do this in a non-magical way.

3) VAST reduction in overall boilerplate (yay!).

tie-rack:
@ddellacosta 

regarding complaint 1:

Handling the connection itself was just added, isn't necessary, and I can get why as a champion of component you don't like it. I can revert that.

regarding complaint 2:

Setting up the functions that call external services could go back to using `async->fn`, with the configuration taking the symbol for the channel passed to `async->fn`. Though, in the tests, you'd still either want to do `with-redefs` (since the tests don't interact with external services) or bring back the weird message routing that was set up at the beginning of each test (which, to me, seems worse).

regarding complaint 3:

Isn't a "bunch of small modular configuration functions for setting up services" what `wire-up` is?

tie-rack:
@ddellacosta I reverted the connection handling and brought `augment-address` back to being defined explicitly via `async->fn`.

Anyone have a better idea for the name than `configure!`?

ddellacosta:
> Setting up the functions that call external services could go back to using async->fn, with the configuration taking the symbol for the channel passed to async->fn. Though, in the tests, you'd still either want to do with-redefs (since the tests don't interact with external services) or bring back the weird message routing that was set up at the beginning of each test (which, to me, seems worse).

Yeah, I don't have a great idea here. Even if we assume that this gets defined implicitly as part of startup within component, it requires you to pass the system around (a la [the earlier version of this](https://github.com/democracyworks/notification-templates/blob/8423982c0cd625e22458a0fbf0d6cb8ed14899ba/src/notification_templates/core.clj#L38) where I tried it out, albeit in a kind of specialized context) and that implies a whole world of potential pain in trying to ensure you always can get a reference to the system--or else you have to store your system in an atom or something somewhere, just pushing your global life-cycle-dependent implicitly-instantiated somewhere else.  It's all ugly, IMHO.

> Isn't a "bunch of small modular configuration functions for setting up services" what wire-up is?

Yes--and my main beef with those is the way args are passed in, and how much unnecessary boilerplate they implicitly require.  I think that simply cleaning those up would give us a ton of benefit regardless.  But in my mind actually calling those and managing initialization and configuration for any given internal/external/event kehaar service would be better managed by individual modular components a la [this approach](https://github.com/democracyworks/kehaar-component/tree/first-version/src/kehaar_component) or similar. That means I can then start up each one of these one-by-one as I need them for testing or otherwise, isolate them when I'm having an issue with one or the other, and introduce new components in a pretty isolated fashion.

> Anyone have a better idea for the name than configure!?

`init!` ?

All my component drum-beating aside, what with removing rabbitmq from the `configure!` (or whatever it ends up being called) will get us 85% of the way toward our goal of removing boilerplate and making kehaar setup easier to manage.  I've just had the experience that a bunch of opaque, ad-hoc `init!` and `shutdown!` functions are not as nice to work with compared to component because of the strict boundaries it enforces on how you configure and control life-cycle for...anything in your app which has a life-cycle.

That's the last sales pitch I'll make for component, at least assuming I can exercise enough restraint to keep shut up about it...

tank157:
So I'm starting to wade in on this now a bit late. And I haven't dug too deeply into the changes yet, just a cursory look and going over the comments, so take this for what it's worth.

I want this to go further. The main problem I see right now is that configuration for queues is still going to live in 1+N places. By which I mean, it'll live in the offering service, but will also live in every service connecting to it. And that's still a big problem, even if we've simplified things down and removed a fair amount of boilerplate. I don't know how to get around this necessarily, not without making projects dependent upon each other in some ways that seem antithetical to micro-services or implementing some kind of service lookup mechanism (which would be a bigger headache).

Anyway, I'm going to think on it some more. There really should be a way to derive an external-service connection from an internal-service one. And maybe that's somehow with components, I don't really know.

tie-rack:
Closing in favor of #58, which is this, rebased off current master, and squashed.

PR #55, opened 2016-11-29
Handle geocoding errors at a higher level

danielglauser:
And include the user-id in all log statements.

Replaces [this](https://github.com/democracyworks/user-works/pull/45) PR which contains a historical context. Pivotal card [here](https://www.pivotaltracker.com/n/projects/930636/stories/129862097).


src/user_works/handlers.clj
```diff
@@ -60,27 +60,55 @@
             (str country-ocd-id "/state:"
                  (clojure.string/lower-case state))))))
 
+(defn handle-geocoding-issue [user-id address error]
```
cap10morgan: Can we call this fn something like `log-geocoding-error` instead? `handle` makes me think it takes incoming RabbitMQ messages.

src/user_works/handlers.clj
```diff
@@ -60,27 +60,55 @@
...
+             "error:" error))
+
 (defn-traced geocode-address
+  "Calls address-works to geocode an address. Returns the entire
+  address containing OCD-IDs. If the address fails to geocode it
+  returns a keyword description of the error. The return value looks
+  like {:address {...} :error :some-valid-error-value}. Valid error
+  values are:
+    :timeout
+    :not-found"
```
cap10morgan: 👍

src/user_works/handlers.clj
```diff
@@ -276,3 +304,17 @@
...
+    (augment-address {:registered
+                      {:street "1507 Blake St."
+                       :city "Denver"
+                       :state "CO"
+                       :zip "80218"}}) ([v] v))
+
+  (log/error "Response timeout calling address-works to geocode"
+             "address:" address "user-id:" #_user-id)
+
+  )
```
cap10morgan: We should delete this or turn it into a test.

src/user_works/handlers.clj
```diff
@@ -60,27 +60,55 @@
...
           fallback-divisions (fallback-ocd-id (:state address))]
       (if-not (= :timeout response)
         (if (seq ocd-divisions)
-          (merge address {:ocd-divisions ocd-divisions})
-          (merge address {:ocd-divisions fallback-divisions}))
-        (do (log/error "Response timeout calling address-works to geocode address: " address)
-            (merge address {:ocd-divisions fallback-divisions}))))))
+          {:address (merge address {:ocd-divisions ocd-divisions})}
+          {:address (merge address {:ocd-divisions fallback-divisions})
+           :error :not-found})
```
cap10morgan: I'd be curious what others think about this. We're recovering from this error (in the best way we can; obviously it's not as good as finding the address), so should we still return an `:error` response and log an error (as this code does) or should it do something different? Log a warning and return a `:warning` response?

@tank157, @tie-rack, or @ddellacosta thoughts?

src/user_works/handlers.clj
danielglauser: I removed the comment since I've updated the tests and think we have pretty good coverage.

src/user_works/handlers.clj
tank157: It seems like some systems have "scoring" for this kind of eventuality. A full geocoding would be a 100% score, and then various levels of fallbacks degrade that confidence score. And then consumers could choose what to do with a less than 100% score. Just a thought.

src/user_works/handlers.clj
```diff
@@ -60,27 +60,55 @@
...
+  {:user-id #uuid \"some-valid-uuid\"
+   ...
+   :addresses {:registered {...}
+               :mailing {...}}}
+  This fn will augment the addresses nested within the address map
+  with OCD-IDs if they are available from address-works. It will
+  log any geocoding issues."
   [message]
   (let [update-fn (fn [address-hash]
                     (into {} (for [[k v] address-hash]
```
ddellacosta: Stylistic comment: I find this a bit confusing, and I would be inclined to refactor it with `update-fn` as an separate function at the least.  There may be more you can do with the `into` and `for` but I'm not sure about that...I'm having a hard time seeing why this couldn't also be a `map`, for example, but also don't know why that'd be better...

src/user_works/handlers.clj
ddellacosta: Agree with Troy--but add a warning for anything other than a 100% score?  Or maybe we'll never get to 100% so should avoid the noise?

danielglauser: I feel like the nature of the error is more useful than the score because I don't see us having more options than successful geocoding, `:not-found`, or `:timeout`. If there were more options then great, but I don't think we'll see any. Would love to get this back to the front end so it could make better decisions, however I think that work should come as part of an address autocomplete epic. I don't care if the issue is an error or a warning since the plan is to have a Papertrail search post these issues to a new channel, and give Carmen a user lookup/fixup tool that let's help desk deal with these situations. Collecting them will also help us deal with any edge cases.

cap10morgan: Scoring could be interesting but feels like too big of an architectural shift for this PR.

@danielglauser Do you mind if I take another crack at what this card is calling for? I think I need to get into the code to solidify my thoughts about it. I will push to this branch but am happy to back it out if we end up deciding not to go my route.

danielglauser: @cap10morgan Feel free to take a crack at it. At this point my biggest concern is we've spent a lot of dev time on a fairly simple feature that I hope will get ripped out and completely replaced when we move to autocomplete for addresses.

danielglauser:
@cap10morgan I think that looks good and does clean up the code a bit, thanks for pushing the commit. @ddellacosta does this address your concerns?

cap10morgan:
I added some tests too.

ddellacosta:
@danielglauser Still would like that inner function to be separate, and I am uneasy with using reduce with side-effecting functions, but meh...quibbling at this point. I definitely think this more clear.

cap10morgan:
@ddellacosta Can do on splitting that function out. I didn't originally because it leaves so little behind in the `geocode-user-addresses` fn. But is that really a _problem_? I don't have strong feelings either way, so I'm making the change you requested there.

On the reduce for side-effecting functions aspect, I'm not sure I agree with the premise. This isn't about side effects at all (expect the logging, which is maybe what you're referring to?). This is all arguments and return values. The non-determinism comes from using an external service to derive the return values from the arguments. But otherwise it seems like an OK use of `reduce` to me: I've got an incoming map, I need to apply an operation repeatedly to some parts of it to generate a new map.

PR #56, opened 2016-12-05
:user-id -> :id

danielglauser:
Unfortunately we sometimes call the user's id :user-id and sometimes
it's :id. I would love to change it to :user-id everywhere however
that would be a big change across many of our services. Considering we
may move to namespaced keywords with Clojure Spec :user/id, it feels
like the best course of action right now is to fix the glitch.

Tested in Krakenstein.


test/user_works/handlers_test.clj
```diff
@@ -839,7 +839,8 @@
                                                        test-divisions)})]
         (is (every? (fn [[_ addr]]
                       (= test-divisions (:ocd-divisions addr)))
-                    (:addresses (geocode-user-addresses user))))))
+                    (:addresses (geocode-user-addresses user))))
+        (is (contains? (geocode-user-addresses user) :id))))
```
cap10morgan: This doesn't seem like it belongs under the heading of "geocodes every address." Could we move it to a testing block like "retains user id when provided" or something like that (or another existing one if it seems appropriate there)?

danielglauser:
@cap10morgan I don't like the inconsistency either. The key `:user-id` is only used by the streaming `fn`s and in this case, when a user is created in the entity `ns` and we want to return their id so the entire user can be looked back up and returned. Please keep in mind that the identifier for the returned user is `:id` and from then on everything else refers to it as `:id`.

Of note is with our current workflow `turbovote-web` is the only consumer of `user-works.handlers/create-user` and we never do anything in `user-works.handlers/geocode-user` or use any provide id (`:user-id`, `:id`, or anything else) so I don't understand what looking for either identifier name would do for us unless I changed all the code in this `ns` (beyond the create workflow) to look for either key name. I can do so if that's what you suggest, I'm just concerned that we may move to Clojure Spec sooner than later and we'll just undo whatever I'm about to do.

cap10morgan:
@danielglauser OK then I'm fine w/ just addressing the location of that `contains? ... :id` test (or removing it if you prefer).

PR #58, opened 2016-12-20
Use kehaar 0.8.0's configured/init! to set up queues and handlers

tie-rack:
This is a rebased and squashed version of #54.


src/user_works/handlers.clj
```diff
@@ -202,10 +201,13 @@
     (async/put! events-chan user)
     {:status :created, :user user}))
 
+(def create-user
+  (handle-errors (partial create-user* channels/create-events)))
```
ddellacosta: 👍

I've been using this exact pattern elsewhere (`turbovote-admin-works` and `user-works`) as it makes testing anything with globally-defined channels (or anything) much easier.

src/user_works/queue.clj
```diff
@@ -11,139 +11,15 @@
...
 
 (defn close-resources! [resources]
   (doseq [resource resources]
     (when-not (rmq/closed? resource) (rmq/close resource))))
 
-(defn close-all! [{:keys [connections channels]}]
-  (close-resources! channels)
+(defn close-all! [{:keys [connections kehaar-resources]}]
+  (kehaar/shutdown! kehaar-resources)
   (close-resources! connections))
```
ddellacosta: Look at how many lines of code are gone \o/

test/user_works/handlers_test.clj
```diff
@@ -629,123 +624,112 @@
...
+
+    (testing "handles timeout gracefully"
+      (let [user {:first-name "Grover"
+                  :last-name "Cleveland"
+                  :email "not.that.grover@whitehouse.gov"
+                  :addresses {:registration {:street "1810 S Dahlia St"
+                                             :city "Denver"
+                                             :state "CO"
+                                             :zip "80222"}}}]
+        (with-redefs [user-works.handlers/address-works-timeout 500]
```
ddellacosta: So, not sure how deep down the rabbit hole you want to go with refactorings so consider this purely discussion: seems like this should simply be another argument in a variable arity function declaration vs. a globally side-cause.

test/user_works/handlers_test.clj
```diff
@@ -759,65 +743,49 @@
...
-                         channels/addresses ([[response-ch _]]
-                                             (async/>!! response-ch :timeout))))
-          {:keys [address status]} (geocode-address address)
-          request (async/<!! request-ch)]
-      (assert (not= request ::timeout))
-      (is (= #{"ocd-division/country:us" "ocd-division/country:us/state:co"}
-             (:ocd-divisions address)))
-      (is (= :timeout status))))
+                   :state "CO", :zip "80211"}]
+      (with-redefs [user-works.handlers/augment-address
```
ddellacosta: Seems like we could pass a channel fn into geocode-address to test this, similarly to the `(some-fn ...)` `(some-fn* ... chan)` pattern you've used elsewhere.

Dunno-how-deep-you-want-to-go-down-the-refactoring-rabbit-hole-so-this-is-just-discussion provision applies here as well...

PR #59, opened 2016-12-20
Supahfly cli

danielglauser:
Added support for [tools.cli](https://github.com/clojure/tools.cli) for parsing and validating command line arguments.


dev-src/user_works/geography_breakdown.clj
```diff
@@ -31,19 +33,41 @@
...
+        "TurboVote users."
+        ""
+        "Options:"
+        options-summary
+        ""
+        "Optional Arguments:"
+        "filename   an optional file to write the results to."
+        "           Defaults to stdout."
+        ""]
+       (str/join \newline)))
```
tank157: This bugs me for some reason, not greatly, but it takes more characters than "\n", and I assume some promotion has to go on behind the scenes (since it works with strings too...maybe it demotes strings to char arrays? I can't find the source to find out. It's not a big deal, really, just struck me as odd. But then I learned something new, hadn't really seen `\newline` before.

dev-src/user_works/geography_breakdown.clj
```diff
@@ -31,19 +33,41 @@
...
+
+(defn -main [& args]
+  (let [{:keys [options arguments summary errors]}
+        (cli/parse-opts args cli-opts)]
+
+    (cond
+      (:help options) (exit 0 (usage summary))
+      errors (exit 1 (error-msg errors)))
+
+    (geography-breakdown (connect-to-user-works-db (:datomic-uri options))
```
tank157: Just checking to make sure the optional filename argument was tested (ie, prints to console when the filename isn't specified, writes to file when it is)?

dev-src/user_works/geography_breakdown.clj
danielglauser: It was new to me too (ruthlessly stolen from the tools.cli page). `\newline` is a character where `"\n"` is a string. Both work in this case, and I agree with your taste that `"\n"` looks nicer.

dev-src/user_works/export.clj
```diff
@@ -1,50 +0,0 @@
-(ns user-works.export
```
tank157: Ah, ok, comments didn't come across in the Files Changed view, my bad.

PR #60, opened 2016-12-27
Dashboard stats v.1

ddellacosta:
[Tracker card](https://www.pivotaltracker.com/story/show/135366857)

Related PRs:
https://github.com/democracyworks/turbovote-admin-works/pull/17
https://github.com/democracyworks/turbovote-admin-web/pull/6
https://github.com/democracyworks/turbovote-admin-http-api/pull/15
https://github.com/democracyworks/turbovote-admin-specs/pull/1

(tangentially related: https://github.com/democracyworks/resource-config/pull/8)


project.clj
```diff
@@ -27,18 +27,23 @@
...
                                    :password [:gpg :env/datomic_password]}}
   :uberjar-name "user-works.jar"
-  :profiles {:uberjar {:aot :all}
+  :profiles {:uberjar {:aot :all
+                       :jvm-opts ["-Dclojure.spec.check-asserts=true"]}
              :dev-common {:resource-paths ["dev-resources"]
                           :source-paths ["dev-src"]
-                          :dependencies []}
+                          :dependencies []
+                          :jvm-opts ["-Dclojure.spec.check-asserts=true"]}
```
ddellacosta: Naw, this actually tells clojure.spec to always call any `asserts` calls you add, by default it is off. See a little ways down in this section: http://clojure.org/guides/spec#_using_spec_for_validation

dev-src/user_works/research_counts.clj
```diff
@@ -0,0 +1,117 @@
...
+            [clj-time.core :as t]
+            [clj-time.format :as tf]
+            [clojure.edn :as edn]
+            [clojure.string :as str]
+            [datomic-toolbox.core :as db]
+            [datomic-toolbox.query :as query]
+            [datomic.api :as d]
+            [user-works.entities.user :as u]))
+
+(defn user-works-db []
```
ddellacosta: Yeah...you're right I think. Removed.

project.clj
sduckett: gotcha

src/user_works/entities/reporting.clj
```diff
@@ -0,0 +1,85 @@
...
+            :where [[?u :user/id ?uid]]}
+   :args [:_db-placeholder_]})
+
+(defmulti process-filter (fn [_ [field _]] field))
+(defmulti process-pull (fn [_ field] field))
+
+(defmethod process-filter :default [query _] query)
+(defmethod process-pull :default [query _] query)
+
+(defn insert-term
```
ddellacosta: So after consideration I went back to your original request and just added some tests--I realized it probably doesn't make a whole lot of sense to add docs just for this bit.  In any case, take a look and let me know what you think:

https://github.com/democracyworks/user-works/pull/60/files#diff-058b9669e0800a952294c7aaced67627R129

ddellacosta:
@sduckett Finally I think ready for you to re-review!

src/user_works/entities/reporting.clj
sduckett: Yeah, I'm keen on tests, so thank you. FWIW, I noticed that most of `reporting-test` is written in the form `(is (= expected actual))`, but `inserts-terms-properly` is written in the form `(is (= actual expected))`.

ddellacosta: Whoops, good catch--and in general I think `is` and other stuff expects it in the that format too. Fixed.

PR #65, opened 2017-01-17
G1 GC

cap10morgan:
This turns on the G1 garbage collector in the production profile.

[This line](https://github.com/democracyworks/user-works/blob/master/Dockerfile#L9) and [this line](https://github.com/democracyworks/user-works/blob/master/Dockerfile#L18) of the Dockerfile should already be activating this profile in the uberjars we're shipping to production.


project.clj
```diff
@@ -41,8 +41,7 @@
                                    :username [:gpg :env/datomic_username]
                                    :password [:gpg :env/datomic_password]}}
   :uberjar-name "user-works.jar"
-  :profiles {:uberjar {:aot :all
-                       :jvm-opts ["-Dclojure.spec.check-asserts=true"]}
```
tank157: Might also want this on in dev too, for any testing from a dev repl

PR #69, opened 2017-01-18
Allow config of multiple jvm opts via env/consul

tank157:
Assigning to @cap10morgan


Dockerfile
```diff
@@ -19,6 +19,8 @@ RUN lein with-profiles $env,datomic-repo uberjar
 
 ENV DATOMIC_OBJ_CACHE_MAX=${DATOMIC_OBJ_CACHE_MAX:-256m}
 
-CMD java -XX:+UseG1GC -Ddatomic.objectCacheMax=$DATOMIC_OBJ_CACHE_MAX \
+ENV JVM_OPTS=${JVM_OPTS:-XX:+UseG1GC}
+
+CMD java $JVM_OPTS -Ddatomic.objectCacheMax=$DATOMIC_OBJ_CACHE_MAX \
```
cap10morgan: Should we fold the Datomic objectCacheMax setting into the JVM_OPTS?

docker-compose.yml
```diff
@@ -15,7 +15,9 @@ services:
       USER_WORKS_AWS_S3_BUCKET: "turbovote-admin"
       INITIALIZE_DATOMIC: "true"
       NEW_RELIC_LICENSE_KEY: ''
-  
+      DATOMIC_OBJ_CACHE_MAX: "128m"
```
cap10morgan: Let's change the default in the Dockerfile instead? If it should be lower, let's make it lower. Or do you think it should just be lower in here?

user-works@.service.template
```diff
@@ -34,6 +34,7 @@ ExecStart=/bin/bash -c 'docker run --name ${CONTAINER} \
   --env LEIN_ARGS="with-profiles datomic-repo,production" \
   --env NEW_RELIC_LICENSE_KEY=$(curl -s http://${COREOS_PRIVATE_IPV4}:8500/v1/kv/new-relic/license-key?raw) \
   --env DATOMIC_OBJ_CACHE_MAX=$(curl -s http://${COREOS_PRIVATE_IPV4}:8500/v1/kv/user-works/datomic-obj-cache-max?raw) \
+  --env JVM_OPTS=$(curl -s http://${COREOS_PRIVATE_IPV4}:8500/v1/kv/user-works/jvm-opts?raw) \
```
cap10morgan: Do we need to quote the value returned by the subshell curl command here since it will often have spaces in it?

Dockerfile
tank157: I thought about that too, it's a system property rather than a jvm arg, maybe at some point we'll have a spot to vary them both? For now it seemed ok to leave on its own.

docker-compose.yml
tank157: Nah, I was just testing that changing the value was reflected in the runtime, I'll put it back to the default.

PR #71, opened 2017-01-24
kehaar.jobs; make user-export asynchronous

ddellacosta:
This is a PR for a few things:
- make user-export asynchronous
- leverage the new kehaar.jobs to do so
- as a byproduct of the above, convert the reporting code to also use kehaar.jobs
- as a byproduct of all of this, change the specs to reflect all of this

I'm not super psyched at how much I'm changing at once, but to some extent it couldn't be helped. Please ask me anything.

Other PRs involved in this work:
https://github.com/democracyworks/turbovote-admin-specs/pull/3
https://github.com/democracyworks/turbovote-admin-http-api/pull/17
https://github.com/democracyworks/turbovote-admin-web/pull/12
https://github.com/democracyworks/turbovote-admin-works/pull/22


project.clj
```diff
@@ -18,7 +18,7 @@
                   :exclusions [commons-codec commons-logging]]
                  [ch.qos.logback/logback-classic "1.1.7"]
                  [org.immutant/core "2.1.5"]
-                 [democracyworks/kehaar "0.8.1"
+                 [democracyworks/kehaar "0.8.2-SNAPSHOT"
```
tank157: We'll want to update this to the full release version when it's available.

src/user_works/handlers/reporting.clj
```diff
@@ -1,31 +1,23 @@
...
-              ", with request-id " request-id
-              ", with query-spec " (pr-str query-spec))
-    (->> (reporting/run-query (db/db) query-spec)
-         (assoc-in message [::rspec/query-results ::uspec/user])
-         query-results-fn)
-    {:status :ok}))
+  [{:keys [::request/request-id]
+    {query-spec ::r.user/user} ::reporting/query-spec
+    :as message}]
+  (spec-utils/valid-or-throw ::reporting/query-context message)
```
tank157: I thought this was something specs could do with a native call? Though, I don't really mind if we have something that is much clearer on what's going on in the name.

src/user_works/handlers/reporting.clj
```diff
@@ -1,31 +1,23 @@
...
+  (log/info "Running query in user-works for request " request-id
```
tank157: Let's change this to a DEBUG statement. The general rule is something that happens infrequently, like on startup, is INFO, and things that can happen a lot as a result of user actions are DEBUG.

ddellacosta: Got it, good to know--I think I had these two muddled in terms of what convention we use. Fixed and additionally I'll take a look in the `turbovote-admin-works` and `turbovote-admin-http-api` PRs to make sure I'm following the same convention.

src/user_works/handlers/reporting.clj
ddellacosta: So, there was nothing that did exactly what we want I think--there is more discussion around this here if you are so inclined:

https://github.com/democracyworks/turbovote-admin-http-api/pull/15#discussion_r94172233

I'm definitely open to considering other things, I don't feel like anything is set in stone--we're still figuring out how we want to use clojure.spec it seems.

project.clj
ddellacosta: Yep, definitely--just waiting for that kehaar update then will tweak this, definitely before merging in regardless of approval status.

src/user_works/handlers/reporting.clj
tank157: Yeah, to be clear, I don't have a problem with it at all, was just curious, and that curiosity is now satisfied.

PR #74, opened 2017-02-06
adds timestamp for user-search; logs delete handler requests

ddellacosta:
Support for user-search in https://github.com/democracyworks/turbovote-admin-http-api/pull/19 and https://github.com/democracyworks/turbovote-admin-web/pull/15


danielglauser:
Do we only want `:created-at` and `:updated-at` when we search for a user? Why wouldn't we want to always return these values?

ddellacosta:
I'm confused--what else do you think we should be returning that we're not?  Longer term we need to make sure we're returning everything as specified in the relevant card (https://www.pivotaltracker.com/story/show/114409453) but I think for now it's reasonable to just return at least the subset we're using in the UI (see perhaps https://github.com/democracyworks/turbovote-admin-web/pull/15/files#diff-823b41817634d19c08823300e4ed34baR50).

danielglauser:
@ddellacosta I was questioning if `:created-at` and `:updated-at` should be added to `entity->user` and not just returned when searching for a users.

ddellacosta:
@danielglauser Hmm, not sure this is the right time/place to do that, not that I necessarily disagree with you on the functional level--`entity->user` is used in a lot of places, and I don't there's a need for any of the other places it's used currently to have the timestamps. Furthermore, I can imagine that this is a costly enough operation that it could potentially affect performance (in particular with `read-user`, which I think we use all over the system) if added to `entity->user` everywhere.

I'd rather tweak it at the point we find ourselves adding the timestamps elsewhere that `entity->user` is called, if you don't object--I think this is a bit of an exception at this point.

danielglauser:
@ddellacosta No problem, just felt like I should ask the question.

PR #75, opened 2017-02-08
Streaming user search

tie-rack:
With the streaming improvements from Kehaar 0.10.0, let's move user search to be streaming!

Attendant PRs:
* [electorate-works#50](https://github.com/democracyworks/electorate-works/pull/50)
* [turbovote-admin-works#24](https://github.com/democracyworks/turbovote-admin-works/pull/24)
* [turbovote-admin-http-api#21](https://github.com/democracyworks/turbovote-admin-http-api/pull/21)


src/user_works/entities/user.clj
```diff
@@ -194,7 +194,8 @@
   [search]
   (let [db (db/db)]
     (some->> (search-query db search)
-             (map (comp user-with-ts (partial d/entity db))))))
+             (sequence (comp (map (partial d/entity db))
+                             (map user-with-ts))))))
```
ddellacosta: Why?

tie-rack: Before `user-with-ts` was introduced, this code was a series of `map` calls, which I changed to a single call to `sequence`. The PR that introduced `user-with-ts` changed the series of `map` calls in a different way, to a single call to `map` with the two functions `comp`ed. When resolving the merge conflict, I favored my way. That's all.

ddellacosta: Okay, so if this is essentially equivalent, is this doing something under the hood that I'm not understanding to avoid turning something from O(n) to O(2n)?  I believe I'm the one who changed this from a series of map calls before, explicitly because of that.

tie-rack: `map` with one argument returns a tranducer, which can be `comp`ed together to produce another tranducer that, yeah, is still O(n) when used.

ddellacosta: Gotcha--thanks for the explanation!

PR #78, opened 2017-04-10
remove any querying on non-registered-address users to speed/shrink things up

ddellacosta:


src/user_works/entities/reporting.clj
```diff
@@ -87,22 +76,13 @@
...
-  [addr-map]
-  (->> addr-map
-       :user/address
-       (filter #(= :registered (:address/type %)))
-       seq
-       boolean))
-
 (defn run-query
-  [db {:keys [::u/pull] :as query-spec}]
+  [db {:keys [::u/pull ::u/filter] :as query-spec}]
```
tank157: Doesn't seem like `::u/filter` is being used in this function, do we need it destructured here?

ddellacosta: Ah yeah, that's a remnant of code I had doing date-filtering stuff which I failed to remove. You're 100% right that we don't need it there but presumably I'll be using it soon-ish, so I'm going to be lazy and leave it in for now if you don't mind.

PR #82, opened 2017-05-17
Add Create Session

zonotope:
### What
* Add an optional `:user/create-session` attribute to User entities. 
* The value of this attribute is read in the `create-session` key of the map passed into `user-works.entities.user/create-user` and should be unique for all users. 
* If `create-user` is called multiple times with the same value for the `create-session` key, only the first will make modifications to the database.

### Why
This will help make sure retry attempts of failed creation attempts because of network connectivity problems don't create multiple user entities when a single entity was intended.


src/user_works/entities/user.clj
```diff
@@ -130,14 +132,29 @@
...
-  (let [user-tx (-> user
-                    ->datomic-tx-datom
-                    add-new-user-id
-                    add-db-temp-id)]
-    {:db-after (:db-after @(db/transact [user-tx]))
-     :user-id (:user/id user-tx)}))
+  (let [user-attributes (user-create-attributes user)
+        create-user-tx [:insert-if-missing user-attributes :user/create-session]
+        db-after (:db-after @(db/transact [create-user-tx]))
+        user-id (if-let [create-session (:user/create-session user-attributes)]
```
ddellacosta: I'm having trouble understanding the logic starting here--after this change are we ever _not_ going to have a `create-session` value?

src/user_works/entities/user.clj
```diff
@@ -130,14 +132,29 @@
...
     (concat user-tx-data address-tx-data)))
 
+(defn-traced user-create-attributes [user]
+  (-> user
+      ->datomic-tx-datom
+      add-new-user-id
+      add-db-temp-id))
+
+(defn-traced find-user-id [db create-session]
+  (ffirst (d/q '{:find [?user-id]
```
ddellacosta: It's of dubious benefit here but you could avoid the `ffirst` by doing

```clojure
(first (d/q '{:find [[?user-id]]
...
```

In this case, six of one, half-dozen the other...

test/user_works/entities/user_test.clj
```diff
@@ -221,6 +222,79 @@
...
+            (testing "multiple calls"
+              (let [{:keys [db-after user-id] :as subject}
+                    (do (users/create-user hank-with-session)
+                        (users/create-user hank-with-session)
+                        (users/create-user hank-with-session)
+                        (users/create-user hank-with-session))]
+                (is (= (count (users/all-users db-after)) 1)
+                    "creates one user")))))))))
+
+(deftest create-user-no-create-session-test
```
ddellacosta: I guess you answer my question above through the existence of this test--so you're anticipating we'll still want to create users without the `session-id` mechanism?  I wonder if we should just scrap it entirely and always assume a `session-id` on create...?

src/user_works/entities/user.clj
zonotope: 👍

test/user_works/entities/user_test.clj
zonotope: i wanted to keep the create session optional so we could deploy the fixes independently since they are from two separate services. i also anticipated other processes that would create users where the create session would not be necessary, like a bulk process for example. i'm fine with just making it a required attribute if you feel like that's a better approach, but i think leaving it optional is more flexible.

cap10morgan: I am 👍 on keeping it optional. In general we should not tie these APIs too tightly to the specific ways in which turbovote-web uses them. Part of the motivation of making the TV backend all separate APIs is to facilitate API-level partner interactions, and some of those may not need this parameter.

test/user_works/entities/user_test.clj
```diff
@@ -127,12 +128,17 @@
     (doseq [u [user1 user2 user3]] (users/create-user u))
 
     (testing "retrieves users by hostnames"
-      (is (= [user1 user3] (->> ["test.turbovote.org" "pokemon.turbovote.org"]
-                                (users/find-by-hostnames (dt/db))
-                                format-users-remove-id))))
+      (let [retrieved-users (->> ["test.turbovote.org" "pokemon.turbovote.org"]
```
zonotope: @ddellacosta i had to change this test and the one below because the old version took order into account. something about this branch made the order come back different. i spent a really long time trying to figure out what change her could cause the returned order to change in this tests, but came up empty. i then took a deep dive into datomic docs and old mailing list posts, and i think i can conclude that datomic doesn't make any guarantees about the sort order of it's result sets, and they expect you to sort stuff in the application.

in light of this, i changed the test to not take order into account. please let me know if order does matter, and if this change will break some other downstream service.

zonotope: @cap10morgan i like your way better. i'll change mine to match yours. hopefully we don't have too much trouble merging.

PR #83, opened 2017-05-24
Convert phone numbers to E.164

zonotope:
This patch adds a `user-works.util.phone-number` namespace that exposes a `conform` function to convert phone numbers to E.164 and a `valid?` function to check if a phone number is valid. Both of these functions make use of google's `libphonenumber` java library.

This patch also uses the `conform` function to convert phone number attributes to E.164 before writing them to the database during create and update actions.


test/user_works/handlers_test.clj
```diff
@@ -551,7 +564,7 @@
       (is (re-find #"\:id missing-required-key"
                    (get-in response [:error :message])))))
   (testing "creating a user then reading it back works"
-    (let [user {:first-name "Eric"
+    (let [user-attrs {:first-name "Eric"
```
gws: Small spacing change here.

test/user_works/handlers_test.clj
```diff
@@ -565,7 +578,8 @@
                                               "ocd-division/country:us"}}}
                 :race :zulu
                 :dob #inst "1981-07-18"
-                :us-citizen true}
+                      :us-citizen true}
```
gws: Small spacing change here.

src/user_works/util/phone_number.clj
```diff
@@ -0,0 +1,28 @@
+(ns user-works.util.phone-number
```
gws: Maybe add a `:refer-clojure :exclude ...` here for the use of `format` below?

gws:
Overall this looks awesome. :+1:

src/user_works/util/phone_number.clj
marlabrizel: Alternate suggestion - rename `format` fn to something like `format-phone`

zonotope: good points. i went with `e164-format`.

zonotope:
@ddellacosta, @tank157: @marlabrizel and i put together a migration to format all the historical phone numbers that are already in the database. the migration uses `libphonenumber` to format the existing phone number, then transacts a `:db/add` datum with the same entity id and the formatted number. please take a look [here](https://github.com/democracyworks/user-works/pull/83/files#diff-04c8a6f1900ceb38c0561df74975668d) and let us know what you think.

dev-src/user_works/phone_number_migration.clj
```diff
@@ -0,0 +1,33 @@
...
+            [datomic.api :as d]
+            [datomic-toolbox.core :as db]))
+
+(def txn-batch-size 100)
+
+(defn query-phone-numbers
+  []
+  (d/q '{:find [?e ?phone]
+         :where [[?e :user/phone ?phone]]}
+       (db/db)))
```
ddellacosta: This is fine but in general (especially in this kind of dev-src code where it doesn't matter) I think it's better to explicitly pass in the db connection vs. relying on datomic-toolbox--this may be a point of contention but in general I think the "side-caused mysteriously appearing database connection" is a bad idea and should not be used for any new code unless absolutely necessary.

dev-src/user_works/phone_number_migration.clj
```diff
@@ -0,0 +1,33 @@
...
+
+(defn build-transactions
+  [query-results]
+  (->> query-results
+       (map phone->tx-data)
+       (partition-all txn-batch-size)))
+
+(defn apply-transactions [tx-data-list]
+  (doseq [tx-data tx-data-list]
+    @(d/transact (db/connection) tx-data)))
```
ddellacosta: Based on my experience copying all the emails to the new `:user/email-search` field, you probably want to use `transact-async` here, and I'd also make sure you get back the results of the transactions just in case (so use `for` or something vs. `doseq`).

dev-src/user_works/phone_number_migration.clj
```diff
@@ -0,0 +1,33 @@
...
+
+(defn query-phone-numbers
+  []
+  (d/q '{:find [?e ?phone]
+         :where [[?e :user/phone ?phone]]}
+       (db/db)))
+
+(defn phone->tx-data
+  [[entity-id phone-number]]
+  (when phone-number
```
ddellacosta: Another quibble but: if you are doing `[?e :user/phone ?phone]` in the query to get these numbers, you shouldn't need (what is implicitly) a nil check here. Maybe `blank?` would make more sense? But hopefully we don't have any blank numbers in the database regardless...

dev-src/user_works/phone_number_migration.clj
```diff
@@ -0,0 +1,33 @@
+(ns user-works.phone-number-migration
+  (:require [user-works.util.phone-number :as phone-number]
+            [datomic.api :as d]
+            [datomic-toolbox.core :as db]))
+
+(def txn-batch-size 100)
```
ddellacosta: FYI I did batches of 5K to copy over the emails and I suspect you could go higher. 100 is going to take you a long time.

resources/schemas/010-add-phone-index.edn
```diff
@@ -0,0 +1,3 @@
+[{:db/id :user/phone
```
ddellacosta: I'm really sorry @zonotope --I screwed up and gave my email migration the same number as your phone number migration. Would you mind bumping this to `011`?

src/user_works/util/phone_number.clj
```diff
@@ -0,0 +1,28 @@
...
+  (:import (com.google.i18n.phonenumbers PhoneNumberUtil
+                                         PhoneNumberUtil$PhoneNumberFormat)))
+
+(def ^:private parser (PhoneNumberUtil/getInstance))
```
ddellacosta: So, another style point but I don't see why these and the functions below need to be private--there's no sensitive state management or the like that we want to prevent API users from accessing directly, and making all of these private precludes the possibility that we may want to use these elsewhere at a later date.

resources/schemas/010-add-phone-index.edn
ddellacosta: (D'oh and I see you already renamed this once! Again, my apologies.)

marlabrizel: Not as part of this PR, necessarily (though maybe we could start in on the goodness here), but we should really start naming these migrations the way we do the others, eg `YYYYMMDD-xxx-migration-name.edn`

zonotope: no problem. the price of renaming a file isn't by the electron. this brings up something i've been thinking about with these migration. what about if we used a long timestamp like rails does instead of a 3 digit integer? it's pretty unlikely for 2 people to create a migration file at precisely the same time. 

one problem that could come up is if the migrations weren't applied in the same order on production as the timestamps based on when the branch is merged. do you think that would make a difference?

or we can just rename files before they're merged.

ddellacosta: Yeah mostly I just apologize for making more bullshit busywork for you.

But yeah, agree with @marlabrizel that we should change to a timestamp if that's a thing we can do, although I don't really know what impact that'll have--my only concern is that it becomes harder to see what order migrations were run in without looking at the db, but not sure how much I care about that (I'll add, I have no idea what datomic-toolbox supports either re: migration naming).

marlabrizel: So one solution that @zonotope and I discussed is renaming just this migration here, and then seeing if we can pull dates from commit history and change the rest in a separate PR. Yay/Nay?

src/user_works/util/phone_number.clj
zonotope: i made them private to explicitly prevent api users from accessing them directly, and so that only `valid?` and `conform` are available outside of this ns. 

imho the bigger the external api, the more locked in to the current implementation, the more you have to maintain, and the bigger potential that small, localized code changes break far away code. if there's some code that absolutely needs to access to any of these functions besides `conform` and `valid?` , it should probably be in this namespace anyway.

i tend to make as many functions private as possible to limit a namespace's external api for these reasons, but it definitely does boil down to a matter of taste. i'm happy to make these functions public if you feel strongly about it.

resources/schemas/010-add-phone-index.edn
ddellacosta: Sure I think it's fine to experiment with this here, may as well!

src/user_works/util/phone_number.clj
ddellacosta: Naw, I won't force the issue if you'd rather leave them private--I just fall on the other side and feel like in a language where using pure functions as much as possible and where immutable data is the norm, the notion of "private" is largely an anachronism and should only (maybe, I'm not even convinced of this) be used to hide ugly implementation-specific state-mutating stuff from the API user.

But I'll concede the point that we're unlikely to use any of these other than here for now, and if we do we'll probably break this out of `user-works` at that point and would revisit then.

But I do see the possibility that stuff like `e164-format` and `validate` could be useful generally!

resources/schemas/010-add-phone-index.edn
zonotope: we might not need to rename the previous files actually. i'm guessing that the name has to be unique, and in order. the timestamp is just a really big number, so `009`, `010`, `011`, .. are going to be smaller than whatever timestamps we have now, so the keys in the database will probably still be in the same order.

i'd need to dig into `datomic-toolbox` regarding migrations, but i'll bet we could also do some minimal/trivial amount of work to make [ragtime](https://github.com/weavejester/ragtime) to do what we want, and allow us to generate migration files either from the command line with `lein`, or from the repl.

dev-src/user_works/phone_number_migration.clj
```diff
@@ -0,0 +1,33 @@
...
+(defn query-phone-numbers
+  []
+  (d/q '{:find [?e ?phone]
+         :where [[?e :user/phone ?phone]]}
+       (db/db)))
+
+(defn phone->tx-data
+  [[entity-id phone-number]]
+  (when phone-number
+    (let [formatted-phone (phone-number/conform phone-number)]
```
ddellacosta: Hey I just thought of something new--I have no idea but is there any possibility `conform` is going to reject something and return nil or an empty string or something? Do we need to check for that here?

src/user_works/util/phone_number.clj
zonotope: The problem with making `e164-format` or `validate` public is that they take in an object of the phone number class defined in `libphonenumber`. If code outside this namespace calls those functions, then they would have to do something with `libphonenumber` explicitly to get the phone number object to pass in. That's leaking the low level implementation of the phone number parsing abstraction, and it would make it that much harder to switch to another phone number library later on.

The two public functions take strings, and that's it. I think that gives us the most flexibility going forward.

resources/schemas/010-add-phone-index.edn
ddellacosta: I would suggest not trying to integrate ragtime at this point--I'm not saying we shouldn't consider it just that it's a bigger change maybe than we want to tackle in this PR, and we should also probably discuss it with the larger team before making that kind of change.

That said sounds like maybe you were simply just suggesting it as one possibility at some point, which seems totally worth trying.

src/user_works/util/phone_number.clj
tie-rack: This sounds like the story my parents told me about where `impl` namespaces come from. ;)

dev-src/user_works/phone_number_migration.clj
zonotope: If the phone number is invalid, then it just removes the non-numerical characters from the string, and then prepends "+1". so "123*45" get's converted to "+112345". 

It does throw an exception if the phone number is something like "a b c d e". i think in those situations, we should just leave the data as it is. That particular user is probably useless to us anyway. 

We don't want to have to halt execution when exceptions get thrown on data like that though, so I'll add some code to catch and log that exception. Good catch.

resources/schemas/010-add-phone-index.edn
zonotope: Ha! Sorry I wasn't clear. I was talking about ragtime (and switching to timestamps instead of integers more generally) in the context of a Friday project *outside* of this pr.

src/user_works/util/phone_number.clj
ddellacosta: Yeah, I think that's a good point and I can't really disagree when discussing this code in particular, as it stands now.

(Fully digressing here and also talking very generally:) I just don't think we should care that much about that kind of thing in Clojure--we should make it as clear as possible what the types are that are expected and let the programmer using the API do what they want. We should make it clear in the docstring what the purpose is and what the args and return values are.

Again, this is in the abstract so is less relevant for discussing this particular code--I can't really disagree with any of the points you've made and I think you're right. Maybe the more important point I want to make is that if we start from the assumption that functions should be pure and that we should be unafraid of exposing data as much as possible, then we'll probably end up writing code rather differently and the notion of "private" and "public" will fall by the wayside.

I think as far as this bit of code is concerned though, the horse has been flogged well and good...

resources/schemas/010-add-phone-index.edn
ddellacosta: Okay I figured, just wanted to make sure.

dev-src/user_works/phone_number_migration.clj
ddellacosta: Okay cool--yeah mostly I wanted to make sure we avoided creating "bad" transaction data. Sounds like we wouldn't have but glad that mentioning this maybe helped avoid exceptions screwing things up.

src/user_works/util/phone_number.clj
ddellacosta: @tie-rack I totally didn't get that joke/point...tell us more!

dev-src/user_works/phone_number_migration.clj
zonotope: I'm 100% with you here. It definitely is a point of contention, and I was trying to follow the pattern in the user entity namespace. I'm glad both you and @tank157 are on the same side of the fence as me about this stuff.

Also, this is the kind of thing I was talking about a few weeks ago when we talking about whether or not global state is a good idea. All things being equal, I would rather pass a db connection into a function as an argument instead of referencing a global var in some other namespace.

tank157: I wouldn't read a whole lot into any patterns in `user-works`, it's definitely some of the oldest code in the rewrite, we've learned a lot since those days. Some of it maybe has been updated, but I'm sure a lot of it is still pretty bad.

src/user_works/util/phone_number.clj
tie-rack: Some projects express the public/private split via the namespace hierarchy. In core.async, e.g., the public API is is in `clojure.core.async`, while the private implementation details live in namespaces below `clojure.core.async.impl`, and it's convention that tells you to not go relying on the `impl` namespaces in your own code.

ddellacosta: @tie-rack ah gotcha, thanks! Great point to mention as part of this discussion.

resources/schemas/010-add-phone-index.edn
cap10morgan: The timestamp-y approach is the one we should be using everywhere. I don't remember why this project started off on the wrong foot there. But I am +1 to correcting it just as soon as we can. Even in this PR.

cap10morgan: Oh, I was too late wasn't I? ;)

zonotope: just a hair too late, but it's in my friday project queue :)

PR #84, opened 2017-05-29
Add optional dupe check on user create

cap10morgan:
[Pivotal card](https://www.pivotaltracker.com/story/show/144865165)

This will need better email and phone dupe checking queries and possibly other logic added into the new :user/create-unique db fn once those research cards are complete. This just does exact field matching queries as placeholders for now.


cap10morgan:
**NOTE:** This should not be merged until the new email and phone search queries are written and pushed.

ddellacosta:
@cap10morgan FYI holding off on reviewing this until I get the email research done and the phone stuff Ben's been working on is in place--let me know if you want any feedback before that. cc @zonotope in case you want to weigh in here too

resources/schemas/20170529001-add-dupe-check-tx-fn.edn
```diff
@@ -0,0 +1,38 @@
...
+  {:lang "clojure"
+   :params [db users]
+   :code (if (< 1 (count users))
+           (throw
+            (IllegalArgumentException.
+             ":user/create-unique only supports one user in the transaction data"))
+           (let [user (first users)
+                 email-user (when-let [email (:user/email user)]
+                              (first
+                               ;; placeholder query; replace w/ efficient & correct one later
```
ddellacosta: So: now that this is merged in here's what you should do I think--

```clojure
(let [lc-email (clojure.string/lower-case email)]
  (d/q '{:find [[?u ...]]
         :in [$ ?lc-email]
         :where [[?u :user/email-search ?lc-email]]}
     db lc-email))
```

...or some variation thereof. Basically, you should be able to take care of the AVET index on the `:user/email-search` field now that it exists, and it should be super fast. Let me know if you actually need a fulltext search as that'll be slightly more complicated (but doable).

ddellacosta:
@zonotope Yeah, enforcing uniqueness on these fields definitely comes up regularly, and would save us some pain in some ways, but for now--until we have some way for users to update their accounts, and maybe there are other considerations I'm forgetting too--we probably want to leave the uniqueness constraint off.

ddellacosta:
Also recording here for posterity, per Carmen:

> Multiple people in a household
> we forced uniqueness in Classic and people who shared emails/phones couldn’t use it

cap10morgan:
@zonotope Do I need to do anything more to optimize the dupe check tx fn for phone numbers?

resources/schemas/20170529001-add-dupe-check-tx-fn.edn
```diff
@@ -0,0 +1,38 @@
...
+                 email-user (when-let [email (:user/email user)]
+                              (let [lc-email (clojure.string/lower-case email)]
+                                (first
+                                 (d/q '{:find [[?u ...]]
+                                        :in [$ ?email]
+                                        :where [[?u :user/email-search ?email]]}
+                                      db lc-email))))
+                 phone-user (when-let [phone (:user/phone user)]
+                              (first
+                               ;; placeholder query; replace w/ efficient & correct one later
```
ddellacosta: Yeah this should work as is--as opposed to the email search stuff, the phone number search can now simply use the field as is because it has an index on it. So this will work. The only thing I might suggest is to clean the phone number first using `clean-phone` ([here](https://github.com/democracyworks/user-works/blob/55f9f42e440896d96bc221e4fe19aa119082d658/src/user_works/entities/user.clj#L209)).

ddellacosta: @zonotope Can ignore Wes's question, I answered above ^

zonotope: 👍

ddellacosta:
Okay ignore my last comment, looks like we are calling `conform-phone` already on the phone number before the insert attempt, so we are good to go on this I think. LGTM! 👍

cap10morgan:
@zonotope Hmm, the more I look at it, the more I think there's a fair amount of overlap between the create-session stuff and this branch. I think we may want to take a step back and figure out what, if anything, should be common between these two features.

cap10morgan:
I'm working up a new PR that builds off @zonotope's excellent work on the create-session feature to do this better. He and just paired on the hard parts. Now I just need to copy-paste some code around and make sure it all still works.

PR #85, opened 2017-05-31
Strip namespaces from message map keys

cap10morgan:
As we discussed on Slack.

Surely there's a way to eliminate the explicit recursion, but it wasn't jumping out at me.

I think this probably needs more namespace stripping for response messages, but this seemed like a good start.


src/user_works/handlers.clj
```diff
@@ -43,6 +44,38 @@
...
+  values intact in the returned copy. It is intended to allow services to
+  start communicating with namespaced keys while we evolve from schema to
+  clojure.spec. Eventually it should be deleted when we use spec for all
+  validations and every service speaks and understands the namespaced keys."
+  [coll]
+  (specter/transform (specter/walker map-entry?)
+                     (fn [[k v]]
+                       [(denamespace k)
+                        (if (coll? v)
+                          (strip-namespaces-from-keys v)
```
zonotope: right. this is the recursive call you mentioned we could avoid if we strip namespaces from all keywords and not just map keys.

cap10morgan:
@zonotope OK, that's done. Along with newer deps and a test.

cap10morgan:
@zonotope I pushed some new commits that reorganize the code a bit but also strip namespaces from incoming responses to the one external service that user-works uses: address-works. It does this by piping the kehaar response channel through a bespoke channel with a namespace-stripping transducer on it and returning that channel (because kehaar `async->fn` consumers will be expecting to take from a core.async channel). Let me know what you think of that pattern for handling incoming reponsese.

src/user_works/handlers.clj
```diff
@@ -4,14 +4,22 @@
...
             [com.climate.newrelic.trace :refer [defn-traced]]))
 
-(def augment-address (wire-up/async->fn channels/addresses))
+(def augment-address*
+  (wire-up/async->fn channels/addresses))
+
+(defn augment-address [addr]
+  (let [response-channel (async/chan 1 util/strip-namespaces-from-keywords-xf)]
+    (async/pipe (augment-address* addr) response-channel)
+    response-channel))
```
zonotope: this line isn't necessary. it isn't documented, but [async/pipe returns the `to` channel by default](https://github.com/clojure/core.async/blob/master/src/main/clojure/cljs/core/async.cljs#L254)

cap10morgan:
@ddellacosta Any objection to merging this? It is intended to be 100% backwards compatible but allow us to start sending namespaced keywords in our messages to user-works. The TV Classic import code is the first thing to do that (and so needs this merged), but it is also intended to start us down the path of using namespaced keywords everywhere. Ironically it does that by stripping them out as the first backwards-compatible step in that direction. And then steps 2 and 3 involve allowing them and requiring them, respectively.

ddellacosta:
Oh I just saw this. Okay so at first I panicked and I was like "oh no this is exactly the opposite of what I want us to be doing" but now I see that it is the opposite and the first step down the path of all that is right and holy (or something like that).

Uh yeah so anyways looks good if so. I'm guessing that at some point we'll simply remove the stripping code once we're ready?  BTW can you link me to the convo in slack that you and Ben were referencing?

cap10morgan:
@ddellacosta That's because I just mentioned you. :) You didn't miss anything.

And yes, we will remove the stripping code just as soon as we are done with it.

Sadly I cannot link you to the Slack conversation as it was in our DM channel.

ddellacosta:
Well, it's probably not necessary anyhow, I get it. Anyways, take this as my "LGTM"--looks good!

PR #86, opened 2017-06-05
Adds migration for fulltext+AVET indexed email search field

ddellacosta:
See https://www.pivotaltracker.com/story/show/146656079 for context
Related: https://www.pivotaltracker.com/story/show/142989085

This needs to be merged-in and we need to do the lower-casing/copy-over of email data into the new attribute created by this PR before we can merge-in https://github.com/democracyworks/user-works/pull/87


resources/schemas/010-add-multi-indexed-email-search.edn
```diff
@@ -0,0 +1,9 @@
+[
+ {:db/id #db/id[:db.part/db]
+  :db/ident :user/email-search
+  :db/valueType :db.type/string
+  :db/cardinality :db.cardinality/one
+  :db/fulltext true
+  :db/index true
+  :db/doc "Provided to support email searching, both lower-cased (leveraging AVET) and fulltext"
```
tank157: I'm curious how lower-cased searching is going to leverage AVET and not fulltext? That seems counter-intuitive.

ddellacosta:
@zonotope I think I addressed this [here](https://github.com/democracyworks/user-works/pull/87#issuecomment-306569829), but let me know if you have further concerns! It's not too late to change how we're going to do this.

zonotope:
@ddellacosta i didn't see that comment. the rationale does make sense, although i'd quibble with leaving the emails as the user entered, but i don't think that matters much if at all. given we *do* need full text then i'm wholeheartedly 👍 here.

ddellacosta:
@zonotope Okay cool! Yeah re: leaving the emails "as-is," apparently there's a possibility that a provider will interpret the local-part case-sensitively, and so on the off-chance that is the case we decided we want to leave that as-is. I'm guessing in reality it's vanishingly few if any and I suspect you're right but ¯\_(ツ)_/¯

PR #87, opened 2017-06-05
Adds code to support fulltext search and using AVET indexed search field to do performant lower-case email searches

ddellacosta:
See https://www.pivotaltracker.com/story/show/146656079 for context
Related: https://www.pivotaltracker.com/story/show/142989085

Depends on: https://github.com/democracyworks/user-works/pull/86

(Also adds fix for how we do pass config in to kehaar in core/queue code.)


src/user_works/entities/user.clj
```diff
@@ -145,7 +155,7 @@
   [db query query-term hostnames]
   (-> query
       (update :in conj '[?hostnames ...])
-      (update :where #(apply conj '[[?user :user/hostname ?hostnames]] %))
+      (update :where conj '[?user :user/hostname ?hostnames])
```
ddellacosta: For @tank157 : is there some reason we wanted `:user/hostname` first before? It seems to me like now that we'll have an index on the email (and phone) attribute, we'll want that to be first and hostnames to be second--can I get a sanity check from you on this?

src/user_works/entities/user.clj
```diff
@@ -130,10 +130,20 @@
                          (:addresses updated-user))]
     (concat user-tx-data address-tx-data)))
 
+(defn add-email-search-attr
```
zonotope: Out of curiosity, why did we decide to go with the extra search attribute instead of normalizing the `email` attribute directly? Could we write a migration that just lowercases and strips the current `email` attribute values, then add a regular index on `email` and call it a day? Will we ever need to search with a partial email address?

src/user_works/entities/user.clj
```diff
@@ -130,10 +130,20 @@
...
+  [user]
```
zonotope: some destructuring could eliminate a few lines. take it or leave it:

```clojure
(defn add-email-search-attr [{:keys [:user/email] :as user}]
  (if-not (str/blank? email)
    (assoc user :user/email-search (str/lower-case email))
    user))
````

ddellacosta: Yeah, I debated doing that and don't know why I didn't; I like the destructured version better (I generally do).

I've tweaked it--thanks for the nudge!

ddellacosta:
So, re: fulltext search--we want this mostly because as of now, Carmen doesn't have an easy way to find users using some kind of wildcard expression--"ddellacosta*", and that's something that would be handy to have for her (and may have other applications, so to be fair there is a certain amount of "throw this in 'cause it may be useful someday" going on). This in turn dictates that we either need a new field for this, or we have to go through some fairly painful messing around to add it to the current `:user/email` field (because you can't add the `fulltext` index after the fact, we'd have to basically rename the current one, etc.).

As far as why we didn't simply add the AVET index to the current email field, that is because we also needed to have a lower-cased email field that would give us fast results (so we can search reliably and efficiently on full email addresses), but wanted to keep the "as the user entered" email field as-is.

These two things together meant that the simplest option was to just add a brand new field, with emails lower-cased by default, with both the AVET and fulltext indexes in place.

src/user_works/entities/user.clj
tank157: Nah, it was just there for optimization. So yeah, if email/phone are in the searches, they should probably go first.

ddellacosta: Cool, thanks!

PR #90, opened 2017-06-12
plumatic/schema -> clojure/spec

zonotope:
Switch from Schema to Spec in user-works. I've defined specs for fields in both the user and address entities, and I made sure that we validate both incoming message payloads immediately as those messages are handled, and we also validate the full entity before we write to the database to make sure the data stays consistent.

### Open Questions and Possible Next Steps
1. We are checking against specific string error messages to test that validation fails when it should, but I think we should be testing against (and returning) data such as status code (in the form of a keyword) and relevant data values that caused the failure instead. I did not change this behavior here and instead tried to massage spec's `explain-data` output to look like schema's error messages because I didn't know how much downstream consumers of `user-works` relied on the specific error messages. I think we should return a map of meaningful data when validations fail, and rely on the user-facing services that made invalid calls to user-works to format that data in the context of that particular service and what the user expects from it.

2. We are implicitly using global state to write to the database directly in the `user-works.entities.user` ns. This means that those functions are not pure, so they make the tests more brittle and difficult to modify or expand on. This also means that we are complecting our business logic with the particular intricacies of our data storage solution. This makes it much harder to change and reason about our business logic pipeline (like adding a validation step before we write to the database, for example). In order to limit the scope of this pr to just moving to spec, I got around that by leaning on datomic's `with` function to check that an entity was legit by doing a shadow write first, but I'm not ecstatic about that solution, and I think we should try to decomplect the business logic from data storage instead long term. 

3. We are using raw, explicitly crafted data in our tests. If this pr is accepted, we should move to using spec's generators to produce data to test our functions with. That will make the tests cover a wider range of input over time, and make the tests less brittle and easier to work with.


cap10morgan:
Re 1: We _do_ return a status code under the `:status` key currently, and use that (for example) in `user-http-api` to set the HTTP response code. But is there more / something else you'd like to see here?

Re 2: I would love to see an example refactor along these lines at some point. It sounds like a good improvement. Maybe for a future Friday project? :)

Re 3: I am of the opinion that each testing approach is valuable in its own right, and that having both is ideal. I like the straightforwardness of example-based tests, and how well you can tie the test scenarios to things resembling actual production data and then debug things relatively straightforwardly when tests are failing. I like generative tests' ability to ferret out dark corner case bugs, but they are harder to reason about when they fail and can leave important production logic paths untouched in some cases. And, as of June 2017, spec's testing integration is still pretty immature. I have to copy some random person's code just to make them run under `lein test` for example. Though I fully expect that to improve over time. So I would love to _add_ spec generator tests instead of replacing the example-based tests with them.

zonotope:
#### Re. Re 1: 
I do like the status code keyword the `user-works` returns from the handlers, and I would love to just use them if they are sufficient. I got tripped up because the tests are explicitly checking that the string error messages match a regular expression in [a](https://github.com/democracyworks/user-works/blob/6161db7b5493d64dd9b3e4f35febf218d3c39df1/test/user_works/handlers_test.clj#L96-L97) [few](https://github.com/democracyworks/user-works/blob/6161db7b5493d64dd9b3e4f35febf218d3c39df1/test/user_works/handlers_test.clj#L112-L113) [places](https://github.com/democracyworks/user-works/blob/6161db7b5493d64dd9b3e4f35febf218d3c39df1/test/user_works/handlers_test.clj#L564-L565) even after they've already checked the status code, so I thought it was possible that some down stream services that call into `user-works` relies on those strings for some reason. If that isn't the case, then I'd be happy to remove those extra tests so they aren't checking an explicit exception message string. 

If there is another service that relies on the explicit english error messages in `user-works`, then I think it's more flexible and makes for less brittle services and tests if we instead put a particular chunk of data in the exceptions `ex-data` slot and check against that, while letting downstream services take that data and format their error messages from it as they please. I think we should favor data over english error message strings because this is an internal service that only communicates with other services and not people. 

#### Re. Re. 2: 
I'm way ahead of you. It's already in the queue. 

#### Re. Re. 3: 
I totally agree with you on the generative testing approach, and I think we're on exactly the same page based on what you wrote, but I wasn't clear before about what I wanted to do re these particular tests. I don't want to replace our tests with generative tests; I want to keep the same tests, but replace our test *data* with data produced by generators. 

We have a lot of [raw maps](https://github.com/democracyworks/user-works/blob/6161db7b5493d64dd9b3e4f35febf218d3c39df1/test/user_works/entities/user_test.clj#L100-L129) repeated over and over in our tests that serve as input in our example based tests. I want to keep the same example based tests, but just change the raw, hard-coded test data to test data produced by the generators. That would make for less brittle tests because when we update the shape/type of the data we expect to take in and make the corresponding changes to the spec (like we would have any way), we won't have to go around manually updating all this test data. The tests will still pass as long as we didn't break anything else. I was thinking of changing the code linked above to something like:
```clojure
(let [user1 (test-user)
      user2 (test-user)
      user3 (test-user)]
...
```
where `test-user` is the generator defined in some test util ns.

src/user_works/entities/address.clj
```diff
@@ -1,10 +1,30 @@
...
+(spec/def :address/state string?)
+(spec/def :address/zip string?)
+(spec/def :address/ocd-division (spec/and string? ocd-division?))
+(spec/def :address/ocd-divisions (spec/coll-of :address/ocd-division))
+
+(spec/def :entity/address (spec/keys :req [:address/street :address/city
+                                             :address/state :address/zip
+                                             :address/type]
+                                       :opt [:address/street-2
+                                             :address/ocd-divisions]))
```
cap10morgan: Looks like some alignment got funky in here.

src/user_works/entities/user.clj
```diff
@@ -84,44 +110,71 @@
...
   (assoc user-datom :db/id (db/tempid)))
 
 (defn- conform-phone [user-attrs]
-  (if (:phone user-attrs)
+  (if (:user/phone user-attrs)
     (try
-      (update user-attrs :phone phone-number/conform)
+      (update user-attrs :user/phone phone-number/conform)
       (catch Exception e
         (log/warn "Got a bad phone number: " (:phone user-attrs)
```
cap10morgan: Need another `:user/phone` edit here?

src/user_works/entities/user.clj
```diff
@@ -84,44 +110,71 @@
...
+(defn add-email-search-attr
+  [{:keys [user/email] :as user}]
+  (if-not (str/blank? email)
+    (assoc user :user/email-search (str/lower-case email))
+    user))
+
+(def default-user-values
+  {:user/language-preference :en})
+
+(defn-traced ->datomic-tx-datom
```
cap10morgan: We should probably `s/datom/data/` here (based on feedback from the de-duping PR).

src/user_works/handlers.clj
```diff
@@ -135,45 +136,18 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(spec/def ::create-attrs (spec/keys :req-un [(or :user/email :user/phone)
+                                             :user/first-name :user/last-name]
+                                    :opt-un [:user/middle-name :user/nickname
+                                             :user/address :user/race :user/dob
+                                             :user/us-citizen :user/hostname
+                                             :user/referral-code :user/party
+                                             :user/language-preference
+                                             :user/partner-comms-opt-in]))
```
cap10morgan: I would be in favor of namespacing incoming messages as the very first thing we do and then validating it w/ `:req` and `:opt`. Any reason that wouldn't work?

src/user_works/handlers.clj
```diff
@@ -317,23 +260,79 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn- all? [val]
+  (= val :all))
+
+(spec/def ::district-division :address/ocd-division)
+(spec/def ::hostnames (spec/or :hostname-list (spec/coll-of :user/hostname)
+                               :wildcard all?))
+(spec/def ::search-email :user/email)
+(spec/def ::search-email-ft :user/email)
```
cap10morgan: Should these two lines be replaced by `(spec/def ::search-email :user/email-search)` now?

src/user_works/util.clj
```diff
@@ -20,3 +30,29 @@
...
+(defn parse-problem [{:keys [pred] :as prob}]
+  (if (seq? pred)
+    (-> pred last first)
+    pred))
+
+(defmulti error-message parse-problem)
+
+(defmethod error-message 'clojure.core/contains? [prob]
+  (let [k (-> prob :pred last last)]
+    (str "Value does not match schema. " k " missing-required-key.")))
```
cap10morgan: I highly doubt we're relying on these error message strings anywhere but user-works' tests. And there it was probably just about making sure it was the expected exception. So I would be 👍 on not duplicating the schema errors anymore and just updating those tests as you see fit.

We'll run the whole thing through a krakenstein before merging just in case, though. :)

test/user_works/handlers/reporting_test.clj
```diff
@@ -1,13 +1,11 @@
...
-    [turbovote-admin-specs.request :as request]
-    [user-works.handlers.reporting :as r]
-    [user-works.test-helpers :as test-helpers]))
+  (:require [clojure.core.async :as async]
+            [clojure.test :refer [deftest is testing use-fixtures run-tests]]
+            [turbovote-admin-specs.reporting :as reporting]
+            [turbovote-admin-specs.reporting.user :as user]
+            [turbovote-admin-specs.request :as request]
+            [user-works.handlers.reporting :as r]
+            [user-works.test-helpers :as test-helpers]))
```
cap10morgan: Don't let @ddellacosta see this! ;)

src/user_works/entities/user.clj
zonotope: 100%. i'm going to change all the incorrect usages of `datom` here.

test/user_works/handlers/reporting_test.clj
zonotope: shhhhhhh, i was trying to sneak it passed him ...

in all seriousness, i think i made this change last week before we talked about how we're going to format ns declarations going forward, so i'll change it back. 

i'm also *extremely* tempted to just write @ddellacosta  an emacs lisp function to sort the entire clojure ns declaration automatically so we can follow like 90% of the oss clojure code i've come across and put the `:require` on the same line as the first dependency. i think stuart sierra is an outlier here, but i'm ok with it either way.

src/user_works/entities/user.clj
```diff
@@ -84,44 +110,71 @@
...
+  [{:keys [addresses] :as user}]
+  (-> user
+      (util/namespace-map "user" user-keys)
+      (->> (merge default-user-values))
```
ddellacosta: Minor style critique: I find this pretty hard to read and think it could be expressed more simply by, say, `(partial merge default-user-values)`. To be blunt, this seems more clever than clear.

Another even more fiddly, minor critique: you don't need the parens around `conform-phone` immediately below.

src/user_works/handlers.clj
ddellacosta: I agree; I'd like to see `:req-un` and `:opt-un` used as little as possible, even if it's about backwards compatibility. I'd much rather see us do a bit of labor now in terms of adding the namespaces on "manually" in a prior step so that we don't have to tweak the `clojure.spec` code once we strip out all the namespace massaging stuff later on.

src/user_works/util.clj
ddellacosta: Agreed; don't think we should need this.

src/user_works/handlers.clj
ddellacosta: Related: is there a reason (assuming we remove `req-un`/`opt-un` that we couldn't just use the `:entity/user` spec from `user-works.entities.user` here?

zonotope: i went back and forth on this and i came down on running validations first because i wanted to (1) distinguish between bugs caused by invalid input and bugs in our namespacing code, and (2) fail fast on invalid input. i can see the value in namespacing first--especially since we plan on sending namespaced keys eventually anyway. i'll push up namespacing to step one.

@ddellacosta you're right. once we push up namespacing, the `:entity/user` spec will work just fine.

src/user_works/entities/user.clj
zonotope: i see what you mean, but i don't think the partial version is particularly clear either. maybe it's just my im**partial**ity to `partial` (get it?).

how about using `as->`, **as in** (damn, i'm on a roll here)
```clojure
(-> user
    (util/namespace-map "user" user-keys)
    (as-> attrs (merge default-user-values attrs)) 
```

that seems clearest to me. do you agree?

ddellacosta: This is still too clever by half, and is simply abusing threading macros, which are intended for helping improve readability of nested function calls. That's not what you have here--you simply need to make sure an argument gets passed to the function in the right position.

So `partial` does this just fine (and in my opinion is the most clear of all the options) but if that doesn't suit you then what about `(fn [attrs] (merge default-user-values attrs))` or `#(merge default-user-values %)`?

I won't insist past this comment, so do whatever floats your boat--but if you're still in doubt I'd suggest asking some other Clojure devs to see what the consensus is here.

ddellacosta: Ugh apparently my memory is terrible and I forgot that you need a double-paren for all of these and I don't know which one of these is worse really.

So as always, everything is terrible. I'm backing out of this, and have fun with your threading macros... :-p

zonotope: i disagree that this is abusing the threading macros because, like i said, i think the `as->` version is the most readable. it's practically english.

if you want, i can drop the thread-first macro and just use `as->` throughout, but i don't see what those extra keystrokes buys us in terms of efficiency or readability (if you already know how `->` works, that is).

zonotope: also, "too clever by half" implies that there are some pitfalls to this approach that i'm missing, and will cause problems later. what are those?

ddellacosta: I won't belabor the point since I think we just have a difference of opinion here--but as I said I think my point ends up being moot because I'm not sure that Clojure gives you anything better to use.

But sure, I guess if I have to choose, I find `as->` more obvious in terms of what is going on than `->>` at least.

I will add that great sins have been committed in programming language history as a result of people trying to make their language or library more like English (or natural languages in general).

But now we're getting into holy wars...haha

test/user_works/handlers/reporting_test.clj
ddellacosta: I don't think we decided on anything, did we? For the record, if I care about anything here I care about alphabetizing more, but if we think it's worth caring about enough to standardize on the team (debatable in and of itself) then we should nail this down in a dev team meeting or vote in #developers, or something where everyone's had a say.

Otherwise I'm happy to stick with keeping the first ns on the same line as `:require` for now. I don't need anyone to write me any emacs lisp macros :-D

src/user_works/entities/user.clj
```diff
@@ -32,11 +34,14 @@
 (spec/def :user/partner-comms-opt-in boolean?)
 
 (spec/def :entity/user
-  (spec/keys :req [:user/id (or :user/email :user/phone) :user/first-name
-                   :user/last-name :user/language-preference]
-             :opt [:user/middle-name :user/nickname :user/address :user/race
-                   :user/dob :user/us-citizen :user/hostname :user/referral-code
-                   :user/party :user/partner-comms-opt-in]))
+  (spec/keys :req [:user/id (or (and :user/email :user/email-search)
```
ddellacosta: `user/email-search` should definitely not be something that is required to enter or be returned--it's really for internal use only. I'm a bit concerned that, if we consider our `specs` to represent the canonical data model for a user, that this will cause some confusion.

I guess it seems to me like the `:user/email-search` field is an implementation detail and should generally not "leak" outside of `user-works` (or get passed in to `user-works`).

zonotope: @ddellacosta i'm using this particular spec to ensure data consistency internally, and to validate our business logic. i think it is part of our business logic that users have an `email-search` attribute that is their email attribute after certain transformations, and any user going to the db with an `email` attribute, but without an `email-search` attribute is inconsistent imo, and has the potential to cause bugs later.

to be clear, i think this is different from validating user input, which i think should be done in the handler when the code is passed in. that's because keys that are valid for creation are different than keys that are valid for updates, and our internal representation is different from the representation the users sees. that's why there are different `::update-attrs` and `::create-attrs` specs in the handler ns. i think we should use this spec to validate all data before we write it to the database, and user the handler specs to validate user input. the handler specs then *should not* validate the `email-search` attribute. 

the validations in the handler might be overkill, but i think it's a good idea to fail fast on obviously invalid input, but i'm not wedded to that. i do think we should be validating everything before it goes to the db however.

ddellacosta: Thanks for your feedback. I actually came around to this perspective discussing it with you in Slack but glad to have your thinking clearly written here as well.

src/user_works/handlers.clj
zonotope: 👍

zonotope:
@cap10morgan @ddellacosta i think i've addressed all of your comments. making the switch to namespace qualified keywords turned this pr into a mammoth though. sorry about that.

src/user_works/entities/address.clj
```diff
@@ -1,31 +1,55 @@
...
+                                                 :address/ocd-divisions]))
+
+(spec/def :entity/address (spec/merge (spec/keys :req [:address/type])
+                                      :entity/address-value))
+
+(def address-keys #{:address/type :address/street :address/street-2
+                    :address/city :address/state :address/zip
+                    :address/ocd-division-id})
+(def internal->external {:address/ocd-division-id :address/ocd-divisions})
+(def external->internal {:address/ocd-divisions :address/ocd-division-id})
```
cap10morgan: Man, that Datomic idiom of singular attribute names even when they are cardinality many really dampens my enthusiasm for just using the full attr names everywhere. I _really_ like plural keys indicating where collection values should be used. 🤔

src/user_works/entities/user.clj
```diff
@@ -297,8 +328,8 @@
   (let [user (entity->user entity)
         timestamps (dbq/timestamps entity)]
     (-> user
-        (assoc :created-at (format-timestamp (:created-at timestamps)))
-        (assoc :updated-at (format-timestamp (:updated-at timestamps))))))
+        (assoc :user/created-at (format-timestamp (:created-at timestamps)))
+        (assoc :user/updated-at (format-timestamp (:updated-at timestamps))))))
```
cap10morgan: Do we maybe want to use a different namespace for these? They aren't really user-specific. What do you think of `:db/created-at` and `:db/updated-at`?

src/user_works/handlers.clj
```diff
@@ -134,55 +136,63 @@
...
+    (-> user
+        (util/denamespace-map-keys)
+        (update :addresses denamespace-address-attrs))))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; create                                                                   ;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(spec/def ::drop-if-duplicate boolean?)
+(spec/def ::create-opts (spec/keys :opt-un [::drop-if-duplicate]))
```
cap10morgan: FWIW, I started calling this `:option/drop-if-duplicate` in the classic importer code.

src/user_works/handlers.clj
```diff
@@ -202,87 +212,168 @@
...
-       :error {:type :not-found}
-       :id id}))
+  [message]
+  (let [id (-> message
+               (namespace-user-attrs)
+               (util/validate ::read-attrs)
+               (:user/id))]
+    (or (some->> id
+                 (user/find-by-id (db/db))
+                 datomic.api/touch
```
cap10morgan: Calling this isn't necessary. I see that it was there before, but we should remove it now. All it does is eagerly load all of the lazily-loaded entity attributes. Handy in a REPL, rarely-if-ever needed in production code.

src/user_works/util.clj
```diff
@@ -19,8 +36,8 @@
 
 (defn denamespace-map
   [mp keys]
-  (-> (into {} mp) ;; turn an entity into a map
-      (namespace-map-keys nil)
+  (-> mp
+      (denamespace-map)
```
cap10morgan: Style thing: We generally omit the parens for fns in thread macros with no other args.

src/user_works/util.clj
```diff
@@ -51,4 +68,15 @@
...
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; validation                                                               ;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn validate [data spec]
+  (when-not (spec/valid? spec data)
+    (let [err-message (spec/explain data)
+          err-data (spec/explain-data spec data)]
+      (throw (ex-info err-message err-data))))
+  data)
```
cap10morgan: This is a nice spec valid-or-throw pattern. 👍

src/user_works/entities/address.clj
zonotope: i totally agree. the first few dives into the code were really confusing because of this. it's good to know that this is a datomic idiom, but it's a really weird one. `:user/addresses` would make so much more sense.

cap10morgan: Yeah, it would. I guess the motivation for the idiom is so that back-references don't look weird or something? Seems like the wrong tradeoff to me, though.

src/user_works/entities/user.clj
zonotope: i agree that these timestamps should not be in the `user` ns because they aren't user specific, but, according to [this](http://docs.datomic.com/schema.html#sec-1-1), the `db` ns reserved for datomic specific concepts like `:db/id` and `:db/cardinality`? do these timestamps count as a datomic specific construct? 

if not, then what about `:timestamp/created` and `:timestamp/updated`?

cap10morgan: Oh, yeah, that's a good point. We probably shouldn't use `:db`. What about `:record/created` or `:entity/created`? I'm trying to capture _what_ was created or updated there.

src/user_works/entities/address.clj
ddellacosta: FWIW this roughly matches with table names in RDBMS--or at least, that's kind of how I think about components in datomic attributes. YMMV

ddellacosta:
Sorry I'm slow to review these changes Ben. I think any remaining issues I have with this code are bigger concerns about how complex a lot of this should-be-simple CRUD code is, which your changes have exposed and helped mitigate rather than exacerbated. Wes's last batch of comments seem on-point to me and more or less the kinds of suggestions I'd make, so I'll leave it at that.

Thanks for all your hard work on this code and patience as you've had to keep merging in new changes! This PR is really moving us in the right direction which is great and makes me happy.

zonotope:
@ddellacosta @cap10morgan I think we also need to do some heavy testing in production before merging this in. As part of these changes, I added new validations after every user update which didn't exist before. I think all data should be validated before writing to the db to ensure consistency, so that's what motivated those changes. 

However, if there is some user data in the production db that doesn't currently pass the validation, and someone tries to update anything about that user, then we won't be able to save the changes because the data won't pass validation. 

Before we merge this, I think it would suffice to run the new validations on all the user records in the production db to make sure they pass. If we do have data that fails validation, we'll need to reconcile it somehow to make it pass.

src/user_works/entities/address.clj
zonotope: it matches table names yes, but it doesn't match the standard RDMS *association* convention. the equivalent of this specific thing in rails/postgres would be a `user` table and an `address` table, and an individual `user` record would have a `has_many :addresses` association, implemented in the db with an `address_ids` column. 

i think it would be a ton more intuitive if the convention here was that a user entity had an attribute called `:user/addresses` with a sequence of individual `address` entities.

src/user_works/entities/user.clj
zonotope: i've been using the `entity` namespace for top-level entity specs, so `:entity/user` and `:entity/address`, so `:entity/created` doesn't fit that pattern. i'll go with `:record/created` for now.

src/user_works/util.clj
```diff
@@ -51,4 +68,15 @@
...
 
-(def strip-namespaces-from-keywords-xf (map strip-namespaces-from-keywords))
\ No newline at end of file
+(def strip-namespaces-from-keywords-xf (map strip-namespaces-from-keywords))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; validation                                                               ;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn validate [data spec]
```
ddellacosta: This probably seems out of the blue BUT just now I realized I needed a function like this so stole it for use in `authorization-works`. However, I think that the args should be reversed to be consistent with how spec passes these in. That is all.

src/user_works/util.clj
```diff
@@ -51,4 +68,15 @@
...
+  (when-not (spec/valid? spec data)
+    (let [err-message (spec/explain spec data)
```
ddellacosta: ...and another thing I just caught here, probably want to wrap this call in `with-out-str`, otherwise this goes to `*out*` by default.

cap10morgan:
@zonotope What's the status of this PR? Should we get it merged in soon? Anything I can help with to make that happen?

zonotope:
this was one of the first friday projects i started working on that ended up being a lot more work than i initially thought. i'll clean it up tomorrow and reassess what it would take to merge it.

PR #92, opened 2017-06-13
turns out we didn't want to strip the namespaces out here!

ddellacosta:


cap10morgan:
Oh! We already had some namespace mojo going eh? Sorry for breaking it. Are there any tests we could add that would have failed on this change?

cap10morgan:
@ddellacosta Saw this got merged, so I wanted to make sure you saw my question above.

ddellacosta:
@cap10morgan yeah sorry, was talking with Ben this whole time but I saw your message. Basically I just missed the fact that the reporting code was working in the fashion we want already, and stripping namespaces actually breaks that.  I'll take a look and see if I can't add some tests to try to detect this but I wanted to get a fix up ASAP because the admin console was broken.

cap10morgan:
@ddellacosta Cool. Yeah, I'm 👍 on getting the fix out quickly. Thanks!

PR #93, opened 2017-06-14
Dupe checking redux

cap10morgan:
Alright, here is the 2nd attempt at optional dupe checking on user create. This time it is using a modified form of the work @zonotope did for the create-session feature (which also about de-duping, so there was some conceptual overlap there).

[Pivotal card](https://www.pivotaltracker.com/story/show/144865165)

I went with `:drop-if-duplicate` instead of `:reject-if-exists` to turn this feature on because it implies less of an error condition. And the way we're doing this, "drop" is a better description of what happens than "reject." Not creating a user because a duplicate already exists isn't an error, it's just a different kind of success (unless we want to be flooded with errors every time we migrate users from TV Classic, which I don't think we do).

I haven't done the performance testing yet, and am kind of feeling like we're at a point where it's more important to get something working out the door even if it's slow.


src/user_works/entities/user.clj
```diff
@@ -158,22 +176,45 @@
...
         user-id (if-let [create-session (:user/create-session user-tx-datom)]
                   (find-user-id db-after create-session)
-                  (:user/id user-tx-datom))]
-    {:db-after db-after
+                  (if created?
+                    (:user/id user-tx-datom)
+                    (-> (find-by-attr-vals
+                         db-after
+                         (select-keys user-tx-datom
+                                      [:user/email-search :user/phone]))
```
zonotope: could we check the `uniq-attrs` keys that were passed in to the function? we *might* want to validate that the `uniq-attrs` are ones we support, but that's probably overkill. i'd just like to avoid hard coding the keys we'll use to find the already existing user, or, at least, add `:user/create-session` to the list here too.

cap10morgan: @zonotope Ah, good catch. What do you think of that most recent commit? Will it screw up on any scenarios?

test/user_works/handlers_test.clj
```diff
@@ -327,6 +327,98 @@
...
+                               {:street "1311 Marion St."
+                                :city "Denver"
+                                :state "CO"
+                                :zip "80218"
+                                :ocd-divisions
+                                #{"ocd-division/country:us/state:co"}}}
+                   :drop-if-duplicate true}
+          create-user (handle-errors (partial create-user* creates-chan))
+          resp1 (create-user message)
+          resp2 (create-user (assoc message :phone "+14055554444"))]
```
ddellacosta: INCREDIBLY minor nitpick but is there any value in adding a phone no. here if we aren't going to test that we do not in fact have a user with this phone no.?

src/user_works/handlers.clj
```diff
@@ -204,13 +205,14 @@
   [events-chan message]
   (log/debug "Received user create message:" (pr-str message))
   (s/validate create-user-schema message)
-  (let [{:keys [user-id db-after]} (-> message
-                                       geocode-user-addresses
-                                       user/create-user)
+  (let [{:keys [user-id db-after status]} (-> message
```
ddellacosta: I'm not sure what I think about this status, and more generally I'm not sure about the status being a value coming from an `entities` namespace unless we take a step back and talk more seriously about these conventions in terms of what we want to do (re: what `handlers` mean vs. `entities` vs. whatever else we may come up with) going forward.

Seems like the appropriate thing to do would be to have the operation throw an exception if we tried to insert something and we did not--then the handler could interpret that however it thought was appropriate. But if we try to create a user and we end up _not_ creating a user then on the level of the meaning of `create-user` that should certainly be an exception, as far as what makes sense to me.

Of course, thinking about that, it also becomes obvious that when we call `create-user` and we get back a user-id but we haven't actually created a user, there is something off. I guess that this function is strictly idempotent as if we pass in `:drop-if-duplicate` then that changes the outcome in a consistent way--so perhaps the issue is naming?  Should this really be `create-or-return-existing-user`?

Or should we really have the functionality to support `:drop-if-duplicate` in a separate handler, and we are trying to avoid some work by complicating the basic functionality of `create-user`?

I'm not 100% sure, but something smells here...just don't know what it is quite yet.

resources/schemas/20170613001-multi-attr-insert-if-missing.edn
```diff
@@ -0,0 +1,23 @@
+[
+ {:db/id #db/id[:db.part/user]
+  :db/ident :insert-if-missing
```
ddellacosta: I'm wondering now if this should be called `insert-if-missing-or-has-unique-attrs` or something like that? Verbose, I know, but it's really not clear what this is doing until you read the actual code.

test/user_works/handlers_test.clj
cap10morgan: The point of changing the phone number there is to ensure that the phone numbers _do_ vary across the two requests and we aren't getting a false positive result just because the code only checks for phone number dupes (or similar).

resources/schemas/20170613001-multi-attr-insert-if-missing.edn
cap10morgan: I think `insert-if-missing` covers it, it's just that it's about the attribute _and_ a given value missing. There's no "this or that" going on here, that's the one thing it does. If there's a name that better conveys that, we could consider it, but this name is already in the production DB (from the create-session work), so the name was kept so that it would replace the old version.

src/user_works/handlers.clj
cap10morgan: That's fair.

In the interest of getting this wrapped up soon, I just pushed a new commit that keeps the roles a bit more separated. It's just reporting on the result of the entity create, and then the handler turns that into a response status.

Does that seem OK as a compromise for now?

src/user_works/entities/user.clj
```diff
@@ -48,6 +48,24 @@
...
+  "Returns all user entities from `db` where any one attribute & value matched
+  one of the key-values supplied in the `attr-vals` map.
+
+  For example, to find all users whose email-search attribute is
+  \"foo@example.com\" OR whose phone number is \"+12223334444\", you would call
+  this fn like this:
+
+  (find-by-attr-vals db {:user/email-search \"foo@example.com\"
+                         :user/phone \"+12223334444\"})"
+  [db attr-vals]
```
marlabrizel: Little nitpick, but does it make more sense to name the second arg `attr-map` instead of `attr-vals`? It took me a moment to realize that this fn expects us to pass a map.

cap10morgan: Hmm, perhaps. Though the docstring does try to make it clear that it should be a map. The name `attr-vals` is intended to imply that it should be a map of attributes to their values. So I think `attr-map` might actually indicate less of its intent?

marlabrizel: Ok, I'm not convinced. I still think that `attr-vals` implies that it's something other than a map, but if nobody else shares this opinion then I'd be willing to let it go. In my mind, `vals` is a partial map (which I don't think even exists, hence the confusion here). Maybe `attrs` would work?

zonotope: i think i agree with @marlabrizel here. my mind is already biased because i worked on this code, but i think if i hadn't then at first glance i'd assume that something called `attr-map` includes both keys and values, while `attr-vals` only contains values.

cap10morgan: OK, I'm happy to change it to make it clearer. I don't love `attr-map` because in general I prefer describing _what_ a data structure holds rather than _which type_ of data structure is doing the holding. So in this case the _what_ is "attributes and their values to find users who match one of them." Would `params` be clearer? As in "search params"?

marlabrizel: Yeah I think `params` is clearer. I'm with you in terms of philosophy; my issue here was just that at first glance, it seems like we actually are trying to describe the actual data structure. `params` or `search-params` (I maybe slightly prefer the latter but don't really care in the end) is fine with me

test/user_works/handlers_test.clj
ddellacosta: Fair enough, that seems reasonable.

src/user_works/entities/user.clj
cap10morgan: But I just PR'd some code where I totally called some things `whatever-map` right at the top, haha. So sure, let's go with `attr-map`. 🤡

zonotope: 👍 for either.

resources/schemas/20170613001-multi-attr-insert-if-missing.edn
ddellacosta: Okay, I guess that my confusion is more that it's not clear what `missing` is referring to, but if you think it works I'm fine with it.

src/user_works/entities/user.clj
ddellacosta: Just saw this. If we want to name it based on what it is vs. what it is made of, maybe `user-attrs` or `user-params`?

resources/schemas/20170613001-multi-attr-insert-if-missing.edn
cap10morgan: Yeah I hear you. I would be up for renaming it if there weren't a bit of a cost involved given the old version. Maybe we don't care? Maybe we would prefer each version having its own name?

resources/schemas/20170613001-multi-attr-insert-if-missing.edn
```diff
@@ -0,0 +1,23 @@
...
+  :db/doc
+  "Add entity attributes only if all of the values of the specified attributes are unique."
+  :db/fn #db/fn {:lang "clojure"
+                 :requires [[datomic.api :as d]]
+                 :params [db datom uniq-attrs]
+                 :code (let [uniq-vals (reduce (fn [m attr]
+                                                 (let [val (get datom attr)]
```
ddellacosta: Sorry, more quibbling about names: [a datom has a pretty distinct meaning](http://docs.datomic.com/glossary.html#sec-14)--but this is a map, as is what is produced [here](https://github.com/democracyworks/user-works/pull/93/files/b9e6122083e546bef09fc7a43820026f079cfcc0#diff-6df3d000a4554bf1c4f6b19231ee55a5R172) and probably there are other uses I'm missing.

resources/schemas/20170613001-multi-attr-insert-if-missing.edn
ddellacosta: I dunno--this bugs me less than other stuff as I can kind of squint and see how it works. I'd push that decision to you and @zonotope , I don't feel too strongly about it now that you've explained it a bit.

src/user_works/handlers.clj
ddellacosta: Honestly I think it warrants a longer conversation and more substantial changes than this, but I'm willing to punt on that for now. None of this is written in stone.

src/user_works/entities/user.clj
ddellacosta: Although, nevermind, it should probably not be `user-*` but something reflecting that this should be about filtering unique vals...anyways, whatever, I don't feel strongly so I'll shut up. 😄

resources/schemas/20170613001-multi-attr-insert-if-missing.edn
cap10morgan: Yeah I had that same suspicion when I inherited the first version of this. Thanks for looking it up. I'll change it.

cap10morgan:
@ddellacosta I'd be happy to talk all this over with you today.

resources/schemas/20170613001-multi-attr-insert-if-missing.edn
zonotope: i don't see the harm in renaming it, as long as we updated the code and remembered not to use the old one. perhaps we could retract the old one from the db? @cap10morgan what are the costs of renaming you mentioned?

cap10morgan: It's probably not a big deal. How does `:transact-if-unique` strike everyone?

zonotope: 👍

PR #94, opened 2017-06-14
Classic import create users

cap10morgan:
[Pivotal card](https://www.pivotaltracker.com/story/show/144867263)

This is the first stage of the classic migration pipeline. Right now it just creates users, as called for by the card. But hopefully it also creates a somewhat sane framework in which to put the rest of the stages too.

I would like to give a walkthrough of this to Team TV soon to go over the basic architecture.

Note: This is based off of the `dupe-checking-redux` branch, so it has those commits too. This PR is only about commit 2ac28b7 (and any that might come after it).


dev-src/user_works/import.clj
```diff
@@ -0,0 +1,111 @@
...
+(def party-map
+  {"Republican"  :republican
+   "Democratic"  :democratic
+   "Green"       :green
+   "Libertarian" :libertarian
+   "None"        :none})
+
+(defn filter-addresses [import-data]
+  (update import-data :user/addresses
+          #(reduce (fn [addrs [addr-type addr]]
```
ddellacosta: Two suggestions:

- split the `fn` inside the reduce out to a separate function called like `filter-addresses*` or something, and
- use `reduce-kv` here.

cap10morgan:
Took your suggestions but split up the address filtering code in a slightly different way (that I better separates the concerns into generalized fns).

I started with HTTP API access because it felt easier than setting up kehaar & RabbitMQ. We can revisit this decision down the road, but I think we need to get the overall classic migration work moving forward soon. And your approval-with-comments is not standing in the way, so thank you!

cap10morgan:
@ddellacosta Before I merge this, I wanted to make sure you were OK with that last commit to parallelize the process. I'm happy to back that out for now if you think it warrants further discussion.

dev-src/user_works/import.clj
```diff
@@ -114,7 +115,12 @@
...
-    (let [all-lines (line-seq reader)]
-      (transduce (importer api-url)
-                 (constantly nil) ; replace with a final report fn?
-                 all-lines))))
+    (let [all-lines (line-seq reader)
+          parallelism 5
+          line-chan (async/chan parallelism)
+          result-chan (async/chan 1000)]
+      (async/onto-chan line-chan all-lines)
+      (async/pipeline-blocking parallelism result-chan (importer api-url)
```
ddellacosta: What does the `pipeline-blocking` block on here, exactly? I'm a bit confused by this point.

ddellacosta:
@cap10morgan Seems good to me, although I'm not 100% grokking the code--see my comment. But yeah, this seems like a cool addition and possible speeding-up mechanism.

dev-src/user_works/import.clj
cap10morgan: It's a confusing name. `pipeline-blocking` isn't blocking on anything itself. There are three pipelining functions in core.async: `pipeline`, `pipeline-blocking`, and `pipeline-async`. The first is for computational parallelism (so doing a lot of work in parallel), the second is for parallelizing work that is blocking (which seemed appropriate here w/ the synchronous API calls we're making), and the third is for parallelizing asynchronous functions (which could also work in this context but is architected pretty differently; for example it takes an asynchronous fn as an arg instead of a transducer). So ultimately for the purpose of parallelizing a transducer that is mostly blocking on synchronous API calls, I chose `pipeline-blocking` as most appropriate.

The docstrings for those fns also get into this a bit (although with not as much as detail as I would ideally like; I'm planning to read through the source for those 3 fns today to make sure I'm not misunderstanding anything).

PR #96, opened 2017-06-26
Create election mail subscription during import

marlabrizel:
This PR is part of the ongoing TurboVote Classic -> 2.0 user migration work.

Specifically, this piece of the puzzle creates an election mail subscription for any user associated with a partner with an `:unlimited` mailing strategy. Users who were not set up to receive paper mailings in Classic will be imported without a new subscription.
This is done by hitting a new http-api endpoint to query a single partner, which then allows us to get election mail subscription information for a given user.

This PR also covers a bit of opportunistic refactoring to move the initial import data manipulation into its own function.

Tested in Krakenstein.

Addresses [this Pivotal Card](https://www.pivotaltracker.com/story/show/144876025)
[Related PR](https://github.com/democracyworks/partner-http-api/pull/8) in `partner-http-api`


dev-src/user_works/import.clj
```diff
@@ -79,28 +79,72 @@
...
+          {:status :error
+           :cause user
+           :reason (:body response)}))
+      (catch Exception e
+        {:status :error
+         :cause e
+         :reason (.getMessage e)}))
+    context))
+
+(defn import-election-mail-subscription
```
ddellacosta: Looking at the error handling in this and `import-user` above, seems like there may be an opportunity to DRY this up somehow with a more generic success-checking/error-handling approach?

dev-src/user_works/import.clj
```diff
@@ -79,28 +79,72 @@
...
-          response (http/post api-url {:content-type :edn
-                                       :body (pr-str user)
-                                       :throw-exceptions false})]
-      (case (:status response)
-        200 {:status :already-exists}
-        201 {:status :success}
-        {:status :error
-         :cause user
-         :reason (:body response)}))
+    {:import-data (edn/read-string line)}
```
ddellacosta: do I smell a monad

dev-src/user_works/import.clj
marlabrizel: You're absolutely right - this code is pretty wet. I think the inclination here was to let the pattern emerge a bit more in the next story and determine an approach there. How do you feel about that?

dev-src/user_works/import.clj
ddellacosta: nevermind, I couldn't tell you what a monad actually _is_ so 🤣

PR #97, opened 2017-06-27
Better classic migration results reporting

cap10morgan:
The previous results reporting was just reporting on the last pipeline stage because each stage overwrote the outcome data of the one before it. This fixes that by giving each stage its own map in the forwarded context and adapts the results reporting fn to look at each one.

It also fixes logging and line numbering which were broken by parallelization. You know what they say, if core.async and transducers aren't solving your problems, you're not using enough of them!


dev-src/user_works/import.clj
```diff
@@ -134,57 +168,100 @@
...
-      :already-exists (do (print ". ") (flush))
-      (println "\nERROR on line" line-num "-" (:reason result)
-               "\nCaused by:" (pr-str (:cause result)) "\n")))
-  result)
+(defn report-results
+  "The last pipeline stage that reports the results of what happened while
+  importing each line."
+  [context]
+  (let [line-num (:line-num context)
+        pipeline-stages {:user "U"
```
zonotope: i think this pipeline stages mapping should be more explicit. maybe a top level `def` binding would be more appropriate.

dev-src/user_works/import.clj
```diff
@@ -134,57 +168,100 @@
...
+  pipeline's transduction. The report data will be a map of success,
+  already-exists, and error counts like the following:
+
+  {:success 42, :already-exists 6, :error 1}"
+  [results-chan]
+  (<!! (async/reduce
+        (fn [report context]
+          (let [result (dissoc context :import-data :line-num)
+                result-key (cond
+                             (every? (fn [[_ {:keys [status]}]]
```
zonotope: i don't feel super strongly about this since this code isn't permanent, but it might help clarity to break out some of these functions into their own `defn` declarations. we're at least three `fn`s deep here.

dev-src/user_works/import.clj
cap10morgan: Good suggestion. I may try to clean this up even more by decoupling this from the shape of the pipeline and just reading it out of the context data structure instead. But if that doesn't pan out, I will at least take your suggestion here.

dev-src/user_works/import.clj
cap10morgan: Yes, that's a good suggestion. Will do.

cap10morgan:
@zonotope Thanks for the review. Let me know what you think of the latest changes.

PR #99, opened 2017-06-29
Import voter registration

zonotope:
Adds a pipeline stage to import a user's voter registration status from classic to 2.0.

I also made a few refactoring changes to make reporting results, http requests, and detecting errors easier.


dev-src/user_works/import.clj
```diff
@@ -187,30 +245,49 @@
...
-          (assoc-in context [:results :import-voting-method-pref]
-                    {:status :error
-                     :cause import-data
-                     :reason (:body response)})))
+          (success context :import-voting-method-pref)
+          (error context :import-voting-method-pref import-data
+                 (:body response))))
+      context)))
+
+(defn import-voter-registration [api-root-url {:keys [import-data results],
```
marlabrizel: Could we add a docstring here to match the other pipeline fns?

dev-src/user_works/import.clj
```diff
@@ -187,30 +245,49 @@
...
 (defn stage-key->stage-tag
   "Converts keyword identifying a pipeline stage."
   [sk]
   (as-> sk $
-        (name $)
-        (str/split $ #"-")
-        (rest $)
-        (map (fn [s] (-> s first str/upper-case)) $)
-        (str/join $)))
+    (name $)
```
marlabrizel: I think I prefer the other indentation. I'm definitely not a Clojure style expert so would love to see where I should be looking for this info, but I feel like in all of bbatsov's examples, subsequent forms are aligned beneath the first argument passed to the threading macro (vs the 2 space indentation)

cap10morgan:
Let's make sure that #100 is merged before this as it refactors a lot of the same code.

dev-src/user_works/import.clj
zonotope: i didn't realize i made this change. this was probably emacs' auto-indentation at some point, so i'll change it back.

i think the difference in alignment is because the "thread-as" macro behaves differently from all the others; it takes two arguments before the threaded forms instead of one.

dev-src/user_works/import.clj
zonotope: I typically only add docstrings when I think they make the intent of the function more clear. if i followed the pattern of the other pipeline fns, the docstring for this function would be something like "Pipeline stage to import the user's voter registration". given the name of the function is `import-voter-registration`, and it's in the pipeline section, that docstring seems redundant. 

i don't feel strongly about it so i can add it if you still think it's necessary, but that's why i left it off.

zonotope:
@marlabrizel i decided to refactor this because repetition is inherently error prone, and i think it's safer to use an abstraction instead. 

we'll have to write more pipeline stages after this (i think). it adds a mental tax to remember, with each new stage, what the structure of the `context` map is and how we should write to it so that later stages know whether the stage we're writing succeeds or fails. to make my reservations concrete, if we try to `assoc-in` to the `context` maps `[:reslts :import-voting-method-preference]`, that could lead to subtle bugs that will not immediately be obvious and are hard to catch later on. if, however, we fat finger `(succss context ...)`, the compiler will complain, or we'll at least get a runtime error. 

what motivated this is that i was running into errors like this from previous steps in the pipeline (we weren't always returning the context from the previous step before). these changes are not about brevity vs verbosity; they're about correctness.

i think one-off code is when you need to be the most careful about stuff like this because the code hasn't been vetted in the real world yet by definition.  so anything that makes mistakes both obvious and easy to fix is a good thing in my book, but i'm happy to rip this stuff out if y'all disagree.

dev-src/user_works/import.clj
marlabrizel: That makes sense. I added docstrings to my functions as that was the established pattern when I started working in here, so my argument in favor is mainly about being consistent.

marlabrizel:
Good arguments :)

I'm all in favor of minimizing potential points of failure (especially of the silent variety), so those are good points.

I would love it if we could merge [#100](https://github.com/democracyworks/user-works/pull/100) first as I did not realize we'd be doing a larger refactor (to Wes' comment above). Maybe moving forward it might make sense to keep opportunistic refactors confined to their own PRs, how does that sound?

zonotope:
yep, i'm all for merging in #100 first. i'm also the scum of the earth because i'm holding up the review. i'm halfway through my review now.

and i totally agree that refactors like this should go in their own prs. i just dived in after i figured out why i was seeing the problems i was. i'll go the separate/discussion pr route going forward.

cap10morgan:
Since this is for the last card in the classic import epic and the other PRs have all been merged, I think the refactors in here are OK (i.e. no need to split them out into a separate PR; but I do like that as a general practice for times when other devs are working on or near the same code).

zonotope:
@marlabrizel based on your last post here, it seems like you're ok with me merging this branch. is that true? if so, could you approve the pull request?

PR #100, opened 2017-06-29
Import & create election notification subscription

marlabrizel:
Imports and creates a user's election notification subscription preference(s), should they have them set. Creates nothing if there are no existing subscriptions.

Tested in Krakenstein

Addresses [this Pivotal card](https://www.pivotaltracker.com/story/show/144879083)


dev-src/user_works/import.clj
```diff
@@ -178,28 +178,63 @@
...
+(defn import-election-notification-subscription
+  "Pipeline state to activate a user's election notification subscription(s)
+   if they exist."
+  [api-root-url {:keys [import-data results] :as context}]
+  (let [user (:import-user results)]
+    (if (and import-data (= :success (:status user)))
+      (if-let [election-noty-subscriptions (:notification/medium import-data)]
+        (let [user-id (get-in user [:user :id])]
+          (reduce
+            (fn [c ens]
```
zonotope: this anonymous function is also fairly hefty. i think the code would be clearer and more scannable if you broke it out into it's own independent function with a clear name and clear argument names. by the time i get to the end of the anonymous function, i've forgotten that it's nested in a reduce, or what's actually being reduced.

dev-src/user_works/import.clj
```diff
@@ -178,28 +178,63 @@
           context))
       context)))
 
+(defn import-election-notification-subscription
+  "Pipeline state to activate a user's election notification subscription(s)
+   if they exist."
+  [api-root-url {:keys [import-data results] :as context}]
+  (let [user (:import-user results)]
```
marlabrizel: Sure. I was following the pattern of the other import stages in the namespace here, so if we change it here we should probably change it everywhere else for the sake of being consistent. Would you agree?

dev-src/user_works/import.clj
```diff
@@ -178,28 +178,63 @@
...
+    (if (and import-data (= :success (:status user)))
+      (if-let [election-noty-subscriptions (:notification/medium import-data)]
+        (let [user-id (get-in user [:user :id])]
+          (reduce
+            (fn [c ens]
+              (let [results-key (-> "import-election-noty-sub-"
```
marlabrizel: ok

dev-src/user_works/import.clj
marlabrizel: I agree it's hefty and not ideal. However, I don't think it's super simple to pull it out into its own named function - we'd end up repeating a lot of what we're doing in the top-level function to get the data into the state that we need it, for example, and I'm not completely convinced of the value of doing that in code that we're not planning to use again. It's extremely possible that I'm overlooking something here, but it seems to me that we'd be exchanging tradeoffs. I'm inclined to leave it the way it is but please point out anything I might be missing.

dev-src/user_works/import.clj
marlabrizel: Hm, I don't think I understand. This transformation is creating a keyword that will hold the result of the subscription import. I am having a much harder time understanding what the threading macro above is doing vs the single threading macro. Can you help me understand what I'm missing? I'm wondering whether implementing some of the other refactoring suggestions made the intent of the single thread clearer. Let me know what you think.

dev-src/user_works/import.clj
zonotope: i didn't realize that you were referring to variables in the top level scope from the anonymous function. making that function easier to read would require re-architecting it, and that probably isn't worth it in this case.

dev-src/user_works/import.clj
zonotope: yep, i'm fine with changing the rest for clarity too.

dev-src/user_works/import.clj
gws: Maybe a threading macro is overkill. The `http/put` line just below this could be similarly rewritten with a threading macro, but is plenty clear on its own.

```clojure
(keyword (str "import-election-noty-sub-"
               (name ens)))
```

marlabrizel: @gws is the intent clearer now that I've implemented some of the other refactoring suggestions?

zonotope: you can think of these threading macros as pipelines. you're putting some input in (the first argument), passing it through some transformations (the rest of the arguments), then getting an answer out. 

your desired output here is the keyword to hold the result of the subscription import, but your input is not the string "import-election-noty-sub-". it's the "ens" variable. that's what you're really pushing through this pipeline to get the keyword you want. if you wrote this as an independent function, the argument that function would take would be the `ens` variable, not the static "import-election-noty-sub-" string. if we got rid of the threading macros, the code would look like this:

```clojure
(keyword (str "import-election-noty-sub- (name ens)))
```

see how the `ens` is the most deeply nested thing? it's like that because it's the thing you are transforming. "import-election-noty-sub-" is an implementation detail of this particular pipeline.

i was just asking to make the fact that `ens` is the real input to the pipeline more clear by putting it as the first argument but, in order to do that, you've got to use the thread last macro instead of thread first because of the way this pipeline is structured--the thing you care about is the last argument to all the functions instead of the first. 

this is exactly why we have both `->` and `->>`. they're both meant to make nested function calls clearer, but `->` works better when the input to the pipeline gets passed to the pipeline functions as the first argument, and the `->>` works better when the input to the pipeline gets passed as the last argument. 

using the thread last macro makes the intent clear in my opinion. in english, the thread last macro version is "first take `ens`, then get it's unqualified name, then prepend 'import-election-noty-sub-' to that, then turn that into a keyword.

zonotope: this page is also super helpful if you haven't seen it already: https://clojure.org/guides/threading_macros

gws: I think the intent is clear through the naming of `results-key` and that's independent of any other refactoring in this file. I happen to "visualize the pipeline" the way Ben does, but rather than debate which threading macro to use I was suggesting to just drop them entirely, since they don't seem to be adding to the readability over the same code without them.

zonotope:
@marlabrizel i've approved these changes. feel free to incorporate some/all/none of my suggestions, or not ;).

PR #101, opened 2017-07-08
fix `if-let` syntax

zonotope:
@ddellacosta this is one last pr for the importer. there was some busted `if-let` syntax which was causing some of the imports to fail. we've got a 100% success rate on the test users now.


dev-src/user_works/import.clj
```diff
@@ -245,28 +245,28 @@
...
-                (if (= 200 (:status response))
-                  (assoc-in import-collection [:results results-key]
-                            {:status :success})
-                  (assoc-in import-collection [:results results-key]
-                            {:status :error
-                             :cause import-data
-                             :reason (:body response)}))))
-            context
-            election-noty-subscriptions))
+           (fn [import-collection election-noty-sub]
```
ddellacosta: I realize that every change here is formatting, and it may not matter too much at this point for this code, but seems like this is a very large function that would do well to have, at least, this function split out for clarity's sake.

zonotope:
@ddellacosta i think i made the same refactor suggestion to marla when i reviewed her pull request that introduced this function. we decided to go against it since it's one-off code that uses other contextual information from outside the function declaration.

PR #102, opened 2017-07-14
Expose the :user/status attribute

tie-rack:
This entire time there's been a `:user/status` attribute that hasn't been exposed. Let's bring it out of shadows.


ddellacosta:
@tie-rack But--what does it represent?

cap10morgan:
It was intentionally relegated to the shadows: https://github.com/democracyworks/user-works/commit/5382206ebd489e4d68c91df8c38c48454566be95

But do you have a new use for it?

tie-rack:
My usage of the `status` attribute was for the kind of chase a tabled user needs (e.g., `:tabling-chase-register`). It's possible this attribute is the wrong place for it, or that a new attribute needs to be added somewhere.

There's no rush on this.

ddellacosta:
Yeah I think this has the potential to cause confusion (it certainly confused me)--would you mind renaming this/adding a new attribute which has a name more precisely describing your use of it?

ddellacosta:
Also I just noticed this in passing when reading up on some other Datomic stuff:

http://docs.datomic.com/best-practices.html#never-remove-or-reuse

PR #104, opened 2017-07-17
Query for user counts by hostname

marlabrizel:
Part of a larger set of PRs to allow for leaderboard generation from the admin console, sans dev intervention.

Adds a query to fetch user counts, sorted by turbovote hostname. The start date is intentionally hardcoded as of 1 Jan 2017; future enhancement will make this date based on user input. For now, a hardcoded start date as of the first of the year acheieves feature parity with the formerly manual chore version.

Also adds a handler so that we can pull this query out and send it to the admin components.

Tested in an admin-only Krakenstein.

Related PRs:
https://github.com/democracyworks/turbovote-admin-http-api/pull/38
https://github.com/democracyworks/turbovote-admin-web/pull/38
https://github.com/democracyworks/turbovote-admin-works/pull/35

[Pivotal Card](https://www.pivotaltracker.com/story/show/149007275)


src/user_works/entities/user.clj
```diff
@@ -305,3 +305,15 @@
   [db user-id]
   (when-let [entity (find-by-id db user-id)]
     (user-with-ts entity)))
+
+(defn user-counts-by-hostname
+  "Gets a count of users associated with each hostname.
+   Used for running leaderboards."
+  [db]
+  (->> (d/q '{:find [[(frequencies ?hostname)]]
```
gws: Try `'{:find [(frequencies ?hostname) .] ...}` instead, which will just return a scalar value. Then, you can drop the threading macro and `first`.

src/user_works/handlers.clj
```diff
@@ -243,6 +243,17 @@
       util/strip-namespaces-from-message
       handle-errors))
 
+(defn get-user-counts*
+  [{:keys [date] :or {date #inst "2017-01-01"}}]
+  {:status :ok
+   :body (->> (d/since (db/db) #inst "2017-01-01")
```
gws: Should `#inst "2017-01-01"` instead be using the value passed in (`date`), rather than repeating the hardcoded value?

src/user_works/handlers.clj
```diff
@@ -243,6 +243,17 @@
...
+              user/user-counts-by-hostname (db/db))})
```
gws: So, `d/since` returns a database value. I'd guess we should only have one call to `(db/db)` here toward the beginning since we want to use the filtered database throughout the query. `user/user-counts-by-hostname` is then correctly using the filtered value, but then that result gets passed to `(db/db)` which seems strange. Maybe double-check this threading macro? Let me know if you don't see a problem; I'm happy to chat more about it. It could be a misunderstanding on my part.

src/user_works/handlers.clj
marlabrizel: This is a prematurely optimized fn (and also my bad for just blindly reusing code here), but there's never any instance (at least, not yet) where a date val is actually being passed in. So, for now, we do need a hardcoded value. As I type this out though, I realize that maybe that's not quite what you're asking 🙃

gws: You suspect correctly, my comment is only about the fact that the hardcoded value is listed twice. If we remove it here (where I left the comment) and leave it up in the arg vector, then when/if somebody passes a real date in, everything will Just Work.

src/user_works/handlers.clj
```diff
@@ -243,6 +243,17 @@
       util/strip-namespaces-from-message
       handle-errors))
 
+(defn get-user-counts*
+  [{:keys [date] :or {date #inst "2017-01-01"}}]
```
ddellacosta: Would you add some validation (even if it's a placeholder) and logging here?

marlabrizel:
@gws @ddellacosta lmk how you two feel about these changes, thanks!

src/user_works/handlers.clj
```diff
@@ -243,6 +243,18 @@
       util/strip-namespaces-from-message
       handle-errors))
 
+(defn get-user-counts*
+  [{:keys [date] :or {date #inst "2017-01-01"} :as message}]
+  (log/debug "Received get-user-counts request for: " (pr-str message))
+  {:status :ok
+   :body (->> (d/since (db/db) date)
+              user/user-counts-by-hostname (db/db))})
```
gws: Were you able to figure this threading macro out? The way I read that form, the return value is a database rather than `user/user-counts-by-hostname`, which doesn't seem right to me.

gws:
@marlabrizel The new query looks great!

ddellacosta:
@marlabrizel Yeah, after thinking about it a bit, probably just as well that we don't add new validation now with Ben's spec changes coming down the pike. The log entry looks good, it's just Gordon's last comment that needs addressing now.

PR #107, opened 2017-10-17
add dev ns to import created time data

zonotope:
related to #106 and the migration there. this patch adds an importer that will import `created_at` data encoded in edn attributes from the classic rails app into the `:user/sign-up-time` attribute.


dev-src/user_works/import_classic_created_at.clj
```diff
@@ -0,0 +1,93 @@
...
+(defn hostname-for
+  "Expects `hostname-map` to be a mapping of old hostnames and their
+  migrated counterparts. Returns the migrated hostname for the old
+  hostname `hostname` passed in."
+  [hostname hostname-map]
+  (get hostname-map hostname))
+
+(defn present? [val]
+  (and (not (nil? val))
+       (not= val "")))
```
ddellacosta: You could shrink this up with `(not (clojure.string/blank? val))`.

zonotope:
I have merged this branch with the branch from #106, so I'm closing this pull request with unmerged commits.

PR #108, opened 2017-10-25
set user `:created-at` from the `:user/sign-up-time` attribute

zonotope:
does what it says on the tin


zonotope:
@ddellacosta please take another look

src/user_works/entities/reporting.clj
```diff
@@ -71,28 +65,43 @@
...
+(defn set-date-range [query start-date end-date]
+  (cond-> query
+    (or start-date end-date) (update-in [:query :where]
+                                        conj '[?u :user/sign-up-time ?ts])
+    start-date (set-start-date start-date)
+    end-date (set-end-date end-date)))
+
+(defn build-query
+  [db {:keys [::u/filter ::u/pull] :as query-spec}]
+  (let [{:keys [::u/start-date ::u/end-date]} filter
```
ddellacosta: The `start-date` and `end-date` values are already part of the `filter` terms--is there some reason you didn't just extend `process-filter` instead of writing a new layer to handle the date filtering? I don't see any reason `process-filter` couldn't support the date filter as well, and more simply and with less code.

I also take issue with the name change of `query-spec->query` (to `build-query`) as I think the original name better describes exactly what the function does and should do: take a query-spec--a map with `filter` and `pull` terms--and return a Datalog query form. I agree the db placeholder was ugly, so that's arguably an improvement, but at the same time it would be nice to be able to call this without a db argument and deal exclusively with data structures. In any case, while I'm on the fence about the db arg I would push to revert the name (and I should probably add a docstring too at some point soon...).

src/user_works/entities/reporting.clj
```diff
@@ -71,28 +65,43 @@
...
+      (update-in query [:query :where] conj '[?u :user/id ?uid]))))
 
 (defn run-query
   [db {:keys [::u/pull ::u/filter] :as query-spec}]
-  (let [{:keys [::u/start-date ::u/end-date]} filter
-        db' (set-db-date-range db start-date end-date)
-        result-set (-> (query-spec->query query-spec)
-                       (update :args #(into [db'] (rest %)))
-                       d/query)]
+  (let [result-set (->> query-spec
```
ddellacosta: At this point you could make all of this a single form, like

```clojure
(->> query-spec
     (build-query db)
     d/query
     (mapv #(zipmap (into [::u/uid] pull) %)))
```

~There's also no longer any reason to destructure the pull and filter args I think.~ Nope, guess you need the pull arg at least. Ah well.

src/user_works/entities/reporting.clj
zonotope: i didn't extend `process-filter` because the `start-date` and `end-date` are not independent with respect to the query like all the other filter terms (though i thought about it because it's a really nice use of multimethods). for example, i need to add a `'[?u :user/sign-up-time ?ts]` rule to the `:where` clause when i add the rule specifying the `end-date`, but only if i didn't already add that rule when i specified the `start-date`. 

i see that the `insert-term` function does dedupe rules before adding them, and i hadn't realized that before. i do think that this implementation is more performant because we don't have to construct a set to do an inclusion test, but building the query is probably not the bottleneck here since we're going to hand it off to a database to do some crunching, and the code would be simpler if i extended `process-filter` instead. ok, you've convinced me.

ddellacosta: @zonotope I think you just convinced yourself there 😆

ddellacosta: I will add, I don't think it hurts to make it so that a `start-date` or `end-date` can be added independently, even if 99.9% of the time we're going to be passing those query terms in together. It should still produce valid queries as far as I can see. In turn this helps keep this code as consistent, general, and declarative as possible.

zonotope: yeah, i agree with that. a lot of the motivation behind writing it the way i did is i wanted to make sure you could specify either one, both, or none with respect to `start-date` and `end-date`, and still come up with a valid query, and i wanted to build the query efficiently. 

the fact that we want to be able to specify them independently but they aren't really independent is what accounts for the extra complexity. but the efficiency lost by building the set to check if a rule has already been added is probably negligible since the clauses are so small.

also i changed the name of `query-spec->query` because now it doesn't just take a query spec, but it also takes a db, as you noticed. i think passing in the db simplifies the code, but that's of course subjective. it was hard (for me) to grok what the `(rest ..` was doing in the args `update` call, until i noticed the placeholder. i didn't see a reason why we shouldn't just start with the db in the slot it needs to be in instead of updating the query after the fact because nothing else referenced the `base-query`. if we eventually want to call it in different ways, we can cross that bridge when we get to it, especially since we'll have more info then on how else we'll want to call it. 

so with the signature change, i thought `query-spec->query` became a little misleading because you need the database (or something else that behaves like one) to build a valid query anyway. at first i renamed it `db-query-spec->query`, but that was hideous, so i settled on `build-query`. i think that's a better description of what's happening because you are building a query out of the arguments.

but, we are literally arguing semantics. i'll change it back if you still think `query-spec->query` is a better fit.

PR #109, opened 2017-11-07
expose endpoint for user ids with signup time by hostname

zonotope:


zonotope:
@cap10morgan this branch was previously approved, but i also had to make some changes that add an index and update the leader board query that i'd like another pair of eyes on.

zonotope:
@cap10morgan also, the endpoint in this branch will be consumed by the code in [this `turbovote-admin-web` pr](https://github.com/democracyworks/turbovote-admin-works/pull/37/files)

PR #111, opened 2017-12-05
Index sign up time

zonotope:
This adds an index to the `:user/sign-up-time` attribute. There are also some importer tweaks that have been living on my hard drive that I wanted in version control for posterity.


dev-src/user_works/import.clj
```diff
@@ -151,6 +152,10 @@
       (dissoc :user/voting-method-preference)
       (dissoc :user/registration-status)))
 
+(defn remap-addresses [user]
+  (update user :user/addresses #(rename-keys % {:registration :registered
+                                                :mailing :mailing})))
```
cap10morgan: Is the `:mailing :mailing` needed so that it doesn't get dropped? I would have thought `rename-keys` would just leave unspecified keys intact in the return value, but I'm not sure I've ever used it.

dev-src/user_works/import.clj
```diff
@@ -205,13 +211,13 @@
                    classic-user->user-works-format
                    (assoc :partner/hostname partner-hostname
                           :option/drop-if-duplicate true))
-          response (post (str/join "/" [api-root-url "users"]) user)
+          response (post (str/join "/" [api-root-url "users/"]) user)
```
cap10morgan: Have you found this trailing slash to be necessary with the latest turbovote-http-api code? It tries to ensure that it is totally optional: https://github.com/democracyworks/turbovote-http-api/blob/master/src/turbovote_http_api/service.clj#L29-L37

dev-src/user_works/import_classic_created_at.clj
```diff
@@ -92,4 +75,71 @@
         tally-ch (chan)]
     (tally-transactions tx-ch tally-ch)
     (async/pipeline-blocking conc tx-ch tx-xf tx-data-ch)
-    (read-input-file infile tx-data-ch)))
+    (read-input-file infile tx-data-ch)
+    {:result tally-ch
+     :stop-fn (fn [] (async/close! tx-ch))}))
+
+;;;;;;; sign up time reset for users skipped during migration
```
cap10morgan: So these are all fns you used / will use at the REPL for one-off data fixes in the DB (versus being added to the usual import pipeline)?

dev-src/user_works/import.clj
zonotope: i don't remember trying it at the time of the change i think i included it here just to be as explicit as possible. i just tried it at the repl, and it works just fine without the `:mailing :mailing`. i would remove it, but i don't think we'll be doing any more imports any time soon.

cap10morgan: I think if the code is worth preserving in git then it's probably worth removing that so it doesn't imply to future readers that it was necessary. But not a huge deal either way, so I'll leave the final decision up to you.

dev-src/user_works/import.clj
zonotope: i don't think i've tried it on the latest `turbovote-http-api`, but i do remember that it was necessary at one point.

dev-src/user_works/import_classic_created_at.clj
zonotope: yep. i've already generated the file containing the datoms i'll need to transact with`write-tx-file!` fn. i have to do a few merges in the right order before i transact those datoms, but these are the functions i will use, and have tested.

dev-src/user_works/import.clj
zonotope: 👍

PR #112, opened 2017-12-12
Reapply sign up time

zonotope:
This branch makes two broad changes.
First, it reapplies the changes reverted in https://github.com/democracyworks/user-works/pull/110 to change the user export response to use the `:user/sign-up-time` field instead of datomic's internal time tracking after the data inconsistencies were corrected in other branches

Secondly, it changes the leaderboard query to use the `:user/sign-up-time` field instead of datomic's internal time tracking to determine how many users have signed up for a given hostname after a given start date


src/user_works/entities/user.clj
```diff
@@ -292,47 +314,32 @@
...
-    (user-with-ts entity)))
+    (with-ts entity)))
 
 (defn user-counts-by-hostname
-  "Gets a count of users associated with each hostname.
-   Used for running leaderboards."
-  [db]
+  "Gets a count from `db` of users associated with each hostname who've
+  signed up since the `date` argument. Used for running leaderboards."
+  [db date]
```
cap10morgan: Does this motivate a name change for this fn? Maybe something like `user-counts-by-hostname-since-date`? It seems a bit more specialized than just `user-counts-by-hostname` now.

src/user_works/entities/user.clj
```diff
@@ -292,47 +314,32 @@
...
   (d/q '{:find [(frequencies ?hostname) .]
+         :in [$ ?date]
          :with [?u]
          :where [[?u :user/hostname ?hostname]
+                 [?u :user/sign-up-time ?sign-up-time]
+                 [(<= ?date ?sign-up-time)]
                  [?u :user/address ?addr]
                  [?addr :address/type :registered]]}
-    db))
+       (d/since db date)
```
cap10morgan: Does this take [the "most callers will refer to the database twice" advice](http://docs.datomic.com/filters.html#since) into account?

zonotope: Thanks for pointing out that link. I had not read it before. I don't think the behavior with respect to that link is any different than what we had before because I just moved the `d/since` from the function caller into the function itself. The code before [filtered the db and then passed it into this function](https://github.com/democracyworks/user-works/pull/112/files/eaca685bc768932e9880f511660c74042c8043ff#diff-037ba8b4a11dcedb374e7c6c9c9b91f6L250), while this branch passes in the unfiltered db into this function, and then filters it here.

Also, this query is meant to return new users that have signed up since this date, so I don't think the older db will have any entities that we care about.

In any case, I included the `d/since` here just as an optimization though. It isn't necessary since we are computing the query by the sign up time. I put in the `d/since` so that the query didn't have to scan the entire db, but that optimization probably doesn't amount to much anyway, so I'm ok with taking out the filter all together and just using the whole db.

zonotope:
@cap10morgan please take another look. i've renamed the function and removed db filtering.

PR #114, opened 2018-01-21
Use address.civic-info service

cap10morgan:
Companion PR to [this address-works change](https://github.com/democracyworks/address-works/pull/80).


cap10morgan:
I didn't actually intend for that `search-users` change to sneak in here too, but if it seems OK to let slip by, it seems like a good thing to have and should be backwards-compatible.

gws:
@cap10morgan I'm 👍 on that `search-users` change. This is a nit in the grand scheme of the changes we're trying to make, but while we're adding new queue names that may be with us a long time, is that the naming scheme we're going for? I thought I'd read somewhere that we were trying to keep them to 3 segments. Maybe it doesn't matter.

cap10morgan:
@gws Yeah it's kind of the naming scheme. There's technically supposed to be a verb in there somewhere, but w/ address-works it's always `read`, `lookup`, or whatever. So I left it out. 🤷‍♂️

PR #115, opened 2018-01-22
Import build pipeline

sduckett:
It would be great if all of the services that gov-notifications used were deployed to that k8s cluster when they were merged to master. This PR takes [the existing Buildkite configuration](https://buildkite.com/democracyworks/user-works/settings) and moves it into version control, adding a trigger to deploy to the gov-notifications k8s cluster.

[Pivotal story](https://www.pivotaltracker.com/n/projects/1603613)


katrinamariehh:
Just to make sure I understand--this and the other build pipeline PRs should mimic what I see in BuildKite but then also include an additional step for k8s as well, right?

sduckett:
@katrinamariehh Yes, this should be the same-ish (tvadmin / tvprod / bsprod ordering isn't important, so long as there's a `wait` before the k8s) as what's setup through the Buildkite UI.

PR #117, opened 2018-01-30
Update leaderboard to count 2018 stats

katrinamariehh:
For [this card](https://www.pivotaltracker.com/story/show/153799525) I updated the `user-counts` function in `user-works.handlers` namespace to pull the year of out today's date for calculating the date to use in the query rather than re-doing the hardcoded date.


katrinamariehh:
I can successfully request a leaderboard locally and I'm not seeing any errors or anything popping up in the krakenstein logs, is there any other testing you'd like to see?

katrinamariehh:
I'm realizing also now that I've actually tested it locally that there will need to be changes to turbovote-admin-web to change the language on the leaderboard page about the date.  Checking with Katy on that now.

katrinamariehh:
Also I was maybe not testing the right thing, so my first comment is possibly off -- I'm not getting any errors generating the leaderboard but I'm also not getting any results which I don't think is right.

zonotope:
i just tested it and my suspicions were correct. here's what i did:
1. check out this branch and bring up krakenstein with `user-works` in dev mode (`./krak new user-works && docker-compose build && docker-compose up`)
2. go through the turbovote sign up process on localhost:58888 to get one user in the db
3. generate a leaderboard on the admin dashboard at localhost:58088. the leaderboard generated this time is a blank page, so it didn't count the user i just signed up.
4. bring down `user-works` with `docker-compose down user-works`.
5. made changes to convert the joda time instance to `java.util.Date`.
6. brought `user-works` back up with `docker-compose up -d user-works`
7. generated another leaderboard from the krakenstein admin dashboard. that new leaderboard now has a single line consisting of `localhost 1`, so the user that i signed up is counted now.

what i linked to before is documentation on how to convert the dates using java interop, but there's an easier way. `clj-time` has this conversion built in, and [that's documented here](https://github.com/clj-time/clj-time#clj-timecoerce). the function you'll want to make use of is `clj-time.coerce/to-date`.

i'm happy to walk you through this in a pairing session if any of that isn't clear.

katrinamariehh:
Yep, like I said I realized I was testing the wrong thing -- getting it fixed now.

zonotope:
also, here is more information on the [types of data datomic can use](https://docs.datomic.com/on-prem/schema.html#attributes). everything that goes into datomic must match one of the entries under the `:db/valueType` bullet. `:db.type/instant` is the only allowable date-related value type, and that maps to `java.util.Date` in jvm languages (including clojure).

src/user_works/handlers.clj
```diff
@@ -240,7 +242,7 @@
       handle-errors))
 
 (defn get-user-counts*
-  [{:keys [date] :or {date #inst "2017-01-01"} :as message}]
+  [{:keys [date] :or {date (c/to-date (t/date-time (t/year (t/today)) 1 1))} :as message}]
```
zonotope: could you make use of the threading macro here? this is a bit hard to read as written, and it looks like it also breaks the 80 character rule. the following would be much clearer to me:
```clojure
(-> (t/today)
    t/year
    (t/date-time 1 1)
    c/to-date)
```
you can also break that out into a helper function if you don't want to write all of that in line:

```clojure
(defn- default-start-date []
  (-> (t/today)
      t/year
      (t/date-time 1 1)
      c/to-date)
```
and then call it like this:
```clojure
(defn get-user-counts*
  [{:keys [date] :or {date (default-start-date)}
    :as message}] 
  ...)
```

PR #118, opened 2018-02-10
Changes for social20 cluster

cap10morgan:
Changes are:

- Use `@(d/transact-async ...)` everywhere instead of `@(d/transact ...)` to avoid the latter's timeouts.
- Turn off spec assert checking in the production jar file. Seems like this is the recommended approach performance-wise?
- Update to latest Datomic (and the version running on the prod transactor now).
- Set max heap to half of the cgroup mem limit. Then we set that limit and k8s mem requests so that we get a good amount of heap, a good amount of RAM beyond that, and make sure k8s leaves enough head room so it doesn't get OOMKilled regularly.

Hoping to un-fork this code and land it on master. I'll go through and make sure mem limit setting is well-configured in all the clusters before merging.


cap10morgan:
@gws Fascinating (re: the no usage of `spec/assert`). 🤷‍♂️

PR #119, opened 2018-02-20
Support voter-district re-mapping

katrinamariehh:
For [this card](https://www.pivotaltracker.com/story/show/153569720) I've added a function `re-map-registered-address-by-ocd-id` that takes in an ocd-id and db, searches for `:registered` addresses with that ocd-id assigned to them, and re-runs those addresses against the geocoding in `address-works`, and then either confirms the existing ocd-ids or retract sones that are no longer accurate and assigns any new ones that are generated.

This has been tested locally and works, but I'm still working on tests (but I wanted to get this up ASAP).


src/user_works/handlers.clj
```diff
@@ -84,6 +84,83 @@
...
+
+(defn-traced re-map-registered-address-by-ocd-id
+  "Gets all addresses with a given ocd-id and attempts to re-map them. If the
+  address geocodes, replace the user's existing ocd-id with whatever was
+  returned."
+  [ocd-id db]
+  (let [registered-address-ids (find-addresses-by-ocd-id db ocd-id)]
+    (->> registered-address-ids
+         (map (partial d/entity db))
+         (map d/touch)
```
zonotope: i don't think we want the `touch` here. if some address entities have attributes we don't need then `touch` will needlessly cache them anyway.

src/user_works/handlers.clj
```diff
@@ -429,3 +506,15 @@
   (-> page-users*
       util/strip-namespaces-from-message
       handle-errors))
+
+(defn-traced update-geocoding*
+  [{:keys [ocd-ids] :as message}]
+  (let [db-val (db/db)
+        results (doall (mapcat #(re-map-registered-address-by-ocd-id % db-val) ocd-ids))]
+    ;; TODO put updated user on channels/update-events
```
zonotope: does this TODO still apply? i think it would be a good idea to broadcast the updates. there are probably services listening for these events that would update themselves automatically in response. if that is the case, then we'd have extra manual cleanup to do after remapping the addresses.

theJohnnyBrown: Yep, still relevant. When I looked into it I realized that the code further down the stack doesn't really deal with users as much as address so it will require some additional querying and rearranging of code.

theJohnnyBrown:
@zonotope I think I've addressed all your comments here. Going to hold off on the pipeline solution until it clearly becomes necessary

PR #120, opened 2018-02-21
log user update messages

zonotope:
we weren't logging calls to the update handler, and those logs would have been helpful if they were in there. i made such a pr for such a small change instead of pushing directly to master in case this wasn't just an oversight and there's a reason we weren't logging updates that i'm not considering.


src/user_works/handlers.clj
```diff
@@ -315,6 +315,7 @@
 
 (defn-traced update-user*
   [events-chan {:keys [id] :as user-map}]
+  (log/debug "Received user update message: " (pr-str user-map))
```
cap10morgan: No need for the trailing space after `message:`. The logging fns will add them for you like `println`.

zonotope:
@cap10morgan i've made the change you requested to get rid of the space and will now merge because the build pass and it's such a small change.

PR #122, opened 2018-04-23
FYI user export script for ocd division

theJohnnyBrown:
@mrichards42 @zonotope


dev-src/user_works/export.clj
```diff
@@ -0,0 +1,61 @@
...
+
+(defn find-users-by-ocd-id
+  "Get all users whose :registered addresses are currently assigned a given ocd-id."
+  [db ocd-id]
+  (d/q '{:find [[?user ...]]
+         :in [$ ?ocd-id]
+         :where [[?address :address/ocd-division-id ?ocd-id]
+                 [?address :address/type :registered]
+                 [?user :user/address ?address]]}
+       db ocd-id))
```
mrichards42: Per your comment above about querying multiple dbs, if you prefix an `:in` arg with `$`, datomic knows those are dbs. Then you have to prepend a db to each of the `:where` clauses. So, something like this:
```
(d/q '{:find [(pull $uw ?user [:user/id :user/first-name :user/last-name]) ?medium]
       :in [$uw $enw ?ocd-id]
       :where [[$uw ?address :address/ocd-division-id ?ocd-id]
               [$uw ?address :address/type :registered]
               [$uw ?user :user/address ?address]
               [$uw ?user :user/id ?user-id]
               [$enw ?e :user/id ?user-id]
               [$enw ?e :notification/medium ?medium]]}
     users-db enw-db ocd-id)
```

dev-src/user_works/export.clj
```diff
@@ -0,0 +1,61 @@
...
+;; apparently datomic can query multiple DBs at once, probably better to do that
+;; https://cjohansen.no/querying-across-datomic-databases/
+(defn merge-notification-type [user]
+  (let [id (:user/id user)
+        medium (:notification/medium
+                (dbq/find-one '{:find [?e]
+                                :in [$ ?userid]
+                                :where [[?e :user/id ?userid]]}
+                              enw-db id))]
+    (assoc user :notification/medium medium)))
```
mrichards42: Heads up for future use of this script that users can have multiple subscribed mediums

mrichards42:
This looks great, thanks for sharing!

dev-src/user_works/export.clj
theJohnnyBrown: 🔥

PR #124, opened 2018-05-14
schedule messages for user ready events

theJohnnyBrown:
Creates an [event indicating the user is ready to export](https://www.pivotaltracker.com/story/show/156697443) in anticipation of ActionKit export and other uses which may arise.


src/user_works/handlers.clj
```diff
@@ -288,7 +316,8 @@
         user-entity (user/find-by-id db-after user-id)
         user (user/entity->user user-entity)]
     (when (= :created result)
-      (async/put! events-chan user))
+      (async/put! events-chan user)
+      (schedule-user-ready-event scheduled/schedule-message-faf {:user-id (:id user)}))
```
mrichards42: I don't know if this falls within the scope of the user-ready event, but it would be helpful to have the hostname in the message as well since crm-works will have to make decisions about what to do with the user based on the hostname.

theJohnnyBrown: The pattern I saw in election-notification-works (which responds to scheduled user events in a similar way) is that it gets the user ID from the event and then calls the `user-works.user.read` endpoint to rehydrate the user object. I wouldn't mind including the whole user record here but you'll probably need to reload it anyway in case more data was added.

That said, I assume we always have the hostname so I'll include that at the very least :)

resources/config.edn
```diff
@@ -85,6 +85,17 @@
            :external-services [{:queue "address.civic-info"
                                 :channel user-works.channels/addresses
                                 :response true
+                                :timeout 30000}
+                               {:queue "scheduled-message-works.message.schedule-faf"
```
katrinamariehh: what is `faf` meant to stand for here? this may not need a change, just curios

src/user_works/handlers.clj
```diff
@@ -14,6 +16,32 @@
             [clj-time.core :as t]
             [clj-time.coerce :as c]))
 
+(defn-traced schedule-user-ready-event
+  "Given a :user-id, schedules ready notifications to go in 10 minutes,
+  depending on ."
```
katrinamariehh: incomplete docstring?

theJohnnyBrown: yep, thanks

resources/config.edn
theJohnnyBrown: [fire-and-forget](https://github.com/democracyworks/kehaar#you-want-to-connect-to-an-external-query-response-service-over-rabbitmq)

PR #125, opened 2018-05-15
Allow passing multiple ocd-ids to export

mrichards42:
This changes the export script to take any number of ocd-ids as arguments (with the last argument still being the output csv file)


dev-src/user_works/export.clj
```diff
@@ -37,18 +37,21 @@
...
                  [?user :user/address ?address]]}
-       db ocd-id))
+       db ocd-ids))
 
-(defn -main [& [ocd-id outfile]] ;; e.g. "ocd-division/country:us/state:az/county:yavapai"
-  (let [export-users (find-users-by-ocd-id users-db ocd-id)]
+(defn -main [& args] ;; e.g. "ocd-division/country:us/state:az/county:yavapai"
+  (let [args (vec args)
+        outfile (peek args)
+        ocd-ids (pop args)
```
gws: These three lines might be combined:

```clojure
[ocd-ids outfile] [(butlast args) (last args)]
```

PR #129, opened 2018-10-25
user export with date ranges

li-at-dw:
[This ticket](https://www.pivotaltracker.com/n/projects/930636/stories/161222475)

Make a uid query that searches on start and end dates.


src/user_works/handlers.clj
```diff
@@ -491,15 +493,13 @@
...
 
 (defn-traced user-signups-by-hostnames*
-  [{:keys [hostnames] :as message}]
+  [{:keys [hostnames start-date end-date] :as message}]
   (log/debug "Handling request for user-ids with sign-up-time for:"
              (pr-str message))
   (s/validate user-ids-by-hostnames-schema message)
   {:status :ok
-   :user-signups (if (= hostnames :all)
-                   (user/all-uids-with-signup (db/db))
```
mrichards42: Is this function used anywhere else, or can we get rid of it now?

src/user_works/entities/user.clj
```diff
@@ -104,13 +104,30 @@
...
+(defn-traced find-uids-with-signup-by-hostnames
+  "Find UIDs and signup times, filtering by hostnames, and
+   start and end dates.  If hostnames is :all, there will be
+   no filtering by hostnames.  start-date and end-date are optional
+   as well."
+  [db hostnames start-date end-date]
+   (let [query-form {:find '[?id ?sign-up-time]
+                     :in '[$]
+                     :where '[]}
+         hn-all (not= hostnames :all)
```
mrichards42: This variable name makes me think that it means that we _are_ asking for all hostnames

src/user_works/entities/user.clj
```diff
@@ -104,13 +104,30 @@
...
+         qf-with-dates (cond-> query-form
+                         hn-all (update :in conj '[?hostname ...])
+                         hn-all (update :where conj '[?user :user/hostname ?hostname])
+                         true (update :where conj '[?user :user/sign-up-time ?sign-up-time])
+                         start-date (update :in conj '?start-date)
+                         start-date (update :where conj '[(>= ?sign-up-time ?start-date)])
+                         end-date (update :in conj '?end-date)
+                         end-date (update :where conj '[(< ?sign-up-time ?end-date)])
+                         true (update :where conj '[?user :user/id ?id]))
```
mrichards42: This is pretty slick, I like it! I think we had a conversation about `cond->` a while back and I couldn't come up with a good example of when you might use it, but this is perfect.

I wonder if it might be more readable if the `:in` and `:where` clauses were built separately? And this might be a controversial suggestion, but some vertical alignment might make it more readable too:

```clj
  (let [use-hn (not= hostnames :all)
        in (cond-> '[$]
             use-hn     (conj '[?hostname ...])
             start-date (conj '?start-date)
             ,,,)
        where (cond-> []
                use-hn     (conj '[?user :user/hostname ?hostname])
                true       (conj '[?user :user/sign-up-time ?sign-up-time])
                start-date (conj '[(>= ?sign-up-time ?start-date)])
                ,,,)
        args (cond-> [db]
               use-hn     (conj hostnames)
               start-date (conj start-date)
               ,,,)
        query {:find '[?id ?sign-up-time]
               :in in
               :where where}])
```

Or maybe even inline in the query itself?
```clj
  (let [use-hn (not= hostnames :all)
        query {:find '[?id ?sign-up-time]
               :in (cond-> '[$]
                     ,,,)
               :where (cond-> []
                        ,,,)}])
```

PR #131, opened 2018-11-12
Use mount for lifecycle management

cap10morgan:
...as discussed in dev team meeting


cap10morgan:
I removed the core.async channel closing code because they are defonce’d and that would render them unusable after one `mount/stop`. Do we need to close them when the app is shutting down?

src/user_works/app.clj
```diff
@@ -0,0 +1,20 @@
+(ns user-works.app
```
eggsyntax: In my experience, it's more natural and idiomatic with mount to put each defstate at a fairly low level rather than gathering them together into a single system configuration. As the mount docs put it, "If the whole app is one big application context (or system), cross dependencies with a solid dependency graph is an integral part of the system...But if stateful things are kept lean and low level, dependency graphs are simple and small." That way individual pieces of state automatically reload when you make changes to the relevant code, and you have less worry about circular dependencies.

This situation is a bit unusual, though. `rabbit-resources` is a good fit for what I'm saying; that could go naturally in `user-works.queue`. But the other two are different, because both datomic-toolbox and resource-config already carry some of the user-works state despite being external libraries. That design is less of a good fit once we're using mount, in my opinion.

So in this particular case, I don't know that there's a much better option for `config` and `datomic-conn` without changing how those libraries work (although I would still move `rabbit-resources`). But I wanted to call it out because I wouldn't want to see people use it as a template for bringing mount into our other services. And maybe we should give some thought to creating versions of those libs that let the calling app hold state instead of owning it themselves (and incrementally switch to the new versions when we bring mount into a service).

src/user_works/app.clj
```diff
@@ -0,0 +1,20 @@
...
+  (:require [datomic.api :as d]
+            [datomic-toolbox.core :as db]
+            [mount.core :refer [defstate]]
+            [resource-config.core :as rc]
+            [user-works.queue :as queue]))
+
+(defstate config
+  "The application config defined in resources/config.edn"
+  :start rc/config
+  :stop (rc/reload-config!))
```
eggsyntax: It seems odd to reload the config on stop. Is that just to ensure that it works properly on restart? If so, I think it would be clearer as
```
(defstate ^{:on-reload rc/reload-config!} config
  "The application config defined in resources/config.edn"
  :start rc/config)
```
Or else (maybe better) create an `rc/stop` that clears the memoization but doesn't restart, and call that on `:stop`.

src/user_works/app.clj
cap10morgan: Good suggestion. I'll take it!

I'll look into moving `rabbit-resources` now.

And yes, the approach taken with `datomic-conn` and `config` were meant to demonstrate a low-effort migration to mount to get the lifecycle management ball rolling. We absolutely should move state out of those libs and into mount down the road. I'll add a comment to that effect in there.

In resource-config's case, however, it's less about storing state and more about optimizing performance. There's a memoized fn in there that reads the config out of the `config.edn` resource, and all that `reload-config!` fn does is tell it to kick that out of the memoized cache so it re-reads `config.edn` on the next call to `config`. This is so you can make calls to `config` in your app all over the place w/o concern that you're re-reading and re-parsing a file every time.

src/user_works/app.clj
cap10morgan: That's all `reload-config!` actually does (i.e. clears the memoization), but that seemed like an implementation detail that shouldn't leak out of resource-config's abstraction.

Do you still think it's inappropriate for `:stop`? What advantage does `:on-reload` have there besides the better semantic alignment?

eggsyntax: None :) . Seems fine to me, then, but if it's just clearing memoization, why call it `reload-config!` as opposed to eg `clear-config!`? Pretty minor point, though; seems fine as-is.

cap10morgan: Because naming is hard? ;)

It seemed to make sense as a name where the observed black box behavior was that before calling this, changes to the underlying `config.edn` file weren't picked up. After calling it, they were. `clear-config!` (or other similar variants I half-considered) seemed to imply other behaviors like "`config` will now return nothing for some reason." But I could be convinced to change it down the road, especially if this ends up causing more confusion than it prevents.

eggsyntax: Makes sense now that you say it that way; I'm happy with `reload-config!` :)

cap10morgan:
@eggsyntax Let me know what you think of this. I eliminated the `app` namespace entirely, and overall I think it's an improvement. As you implied, this work will become a template for other mount migrations once it's merged and I think it's a good idea to push it a little further towards idiomatic mount usage. Which hopefully I've done but let me know if not.

Either way, thanks for the nudge!

test/user_works/handlers_test.clj
```diff
@@ -1,22 +1,29 @@
...
+(defn take-scheduled-messages [tests]
+  ;; Take from user.ready scheduling channel so we don't block on that
+  (go-loop []
+    (let [v (<! channels/scheduled-message-schedule-faf)]
+      (when (not= ::stop v)
+        (recur))))
+  (tests)
+  (put! channels/scheduled-message-schedule-faf ::stop))
+
+(use-fixtures :once take-scheduled-messages)
```
cap10morgan: I think this was maybe relying on a bug in the old version of `core.async` that kehaar was bringing in prior to version 1.0.0. Kehaar's `core.async` dep is now marked as `:provided` (as of kehaar 1.0.0) so that we can provide our own (hopefully recent) version of it in apps that use kehaar.

cap10morgan: Or... perhaps more likely, it was relying on the old 1,000 message buffer on all outgoing channels. But I could have _sworn_ the tests were still passing after that change but before the kehaar & core.async version bump. 🤷‍♂️

test/user_works/test_helpers.clj
```diff
@@ -1,11 +1,15 @@
...
-      (db/initialize {:uri uri, :partition :user-works})
-      (tests)
-      (finally (d/delete-database uri)))))
+  (let [uri (str "datomic:mem://user-works-" (UUID/randomUUID))]
+    (-> (mount/except #{#'queue/rabbit-resources})
+        (mount/swap {#'entities/datomic-config
+                     {:uri uri, :partition :user-works}})
+        mount/start)
+    (tests)
+    (mount/stop)))
```
cap10morgan: I ❤️ what mount allows us to do here so much.

eggsyntax:
> @eggsyntax Let me know what you think of this. I eliminated the `app` namespace entirely, and overall I think it's an improvement. 

Yeah, 100% agreed. I thought about suggesting eliminating that ns, in order to avoid giving the impression that mount is about creating a centralized system configuration, but it didn't seem too important given that 2/3 of the state was just representing state held in libs. But I think you're totally right that this is a more idiomatic way to do it, and I really like that you've broken it up into smaller, more atomic chunks.

Just conceptually, `handlers` doesn't seem like a really natural place for `config`, but since that's the only place it's called in user-works, it seems like a reasonable choice to me. One pattern I've used with mount is to rely on the variance in `:source-paths` to load different config for prod, dev, and test. That can cash out into initializing config in different `:main` namespaces for different environments. I haven't looked into the mechanics of how we currently load & merge the different `config.edn` files, though, so not sure whether it's a good fit for us. Just tossing it out there as something to consider.

cap10morgan:
@gws W/r/t the `core.async` comment (not sure why it won't let me reply directly), that is the intended and desired behavior. Using `core.async` directly was a requirement of kehaar anyway, and it wasn't being updated regularly. So this seemed better.

PR #133, opened 2019-01-09
Code for generating ocd-id counts [skip ci]

mrichards42:
[Pivotal chore card](https://www.pivotaltracker.com/story/show/163058868)

This is sort of a "for fun" PR, but if we ever want to generate ocd-id counts later, it could be useful.


eggsyntax:
Seems like an alternate approach might be to write this as a transducer. It would cut memory use some, at least, by not having intermediate collections.

mrichards42:
@eggsyntax I'm having a hard time understanding what that would look like. I could write something like `user-entity->ocd-ids` as a transducer, and it would probably use a little less memory, but I'm not sure that matters all that much the way this is written. The memory problem I'm trying to avoid here is building up the entire result set in memory, which would happen if I used a regular query.

I definitely don't need to flush all the data to the intermediate csv -- the `reduce` could happen right away -- but having the intermediate csv is helpful in case of network hiccups.

eggsyntax:
I'm not all that clear on it myself; my intuitions on transducers are only so-so. I agree that the memory difference wouldn't matter much relative to your implementation; you've already gotten the important win. I was just imagining that you could rewrite `write-ocd-counts` as a transducer, where the pipeline starts with the reader, includes everything currently in the `->>`, and ends with the writer. Something a bit like `process-with-transducers` [here](https://gist.github.com/borkdude/ddc9433c396b1ff43a091cf2901b9dc7#file-text_xform-clj-L28). Mainly it just seemed like it might be an elegant/readable way to do it (modulo transducer familiarity).

Another (unrelated) option might be an implementation that takes an ocd-id, fetches matching users, appends the count to the output file, and then abandons that data to GC and goes on to the next ocd-id. But I think that would probably be slow as hell unless we happen to have just the right indexes ;)

PR #135, opened 2019-03-08
Add a dev-src script for transferring hostnames

mrichards42:
Somewhat related [pivotal card](https://www.pivotaltracker.com/story/show/159023304)

This is a new (infrequent) chore that should be run from a repl. Comments at the end of the file show a couple of examples for how to query and transfer hostnames.


dev-src/user_works/transfer_hostnames.clj
```diff
@@ -0,0 +1,82 @@
...
+         :where [[?e :user/hostname ?hostname]]}
+       $uw hostnames))
+
+;; -- Transactions ------------------------------------------------------------
+
+(defn transfer-hostname [uw-conn user-entity-ids new-hostname]
+  (doseq [batch (take 1 (partition-all 100 user-entity-ids))]
+    (let [tx-data (mapv (fn [e] [:db/add e :user/hostname new-hostname])
+                        batch)]
+      @(d/transact-async uw-conn tx-data))))
```
eggsyntax: I suspect the `(take 1)` is left over from the dev process, I think it'd mess this up.

eggsyntax:
(the only blocking comment is on the `take 1`; the others are just suggestions.

dev-src/user_works/transfer_hostnames.clj
mrichards42: good catch!

dev-src/user_works/transfer_hostnames.clj
```diff
@@ -0,0 +1,82 @@
...
+(defn all-partner-hostnames [$pw]
+  (set
+   (concat
+    (d/q '{:find [[?domain ...]]
+           :where [[_ :partner/domain ?domain]]}
+         $pw)
+    (d/q '{:find [[?domain ...]]
+           :where [[?a :alternate-hostname/hostname ?domain]]}
+         $pw))))
+
```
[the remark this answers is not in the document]
mrichards42: I didn't know you could bind different values to the same variable with `or`. that's much better

dev-src/user_works/transfer_hostnames.clj
```diff
@@ -0,0 +1,82 @@
...
+         :in [$ [?hostname ...]]
+         :where [[?e :user/hostname ?hostname]]}
+       $uw hostnames))
+
+(defn users-by-hostname [$uw hostnames]
+  (d/q '{:find [[?e ...]]
+         :in [$ [?hostname ...]]
+         :where [[?e :user/hostname ?hostname]]}
+       $uw hostnames))
+
```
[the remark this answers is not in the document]
mrichards42: I think `all-user-hostnames` would be a lot slower (slash might run out of memory?) if it had to return all the user ids as well. If I remember correctly, assuming you aren't calling this on hostnames with a ton of users, once you call either `users-by-hostnames` or `counts-by-hostnames` the user data is loaded into datomic's cache, so the next call to either fn is pretty fast.

dev-src/user_works/transfer_hostnames.clj
```diff
@@ -0,0 +1,79 @@
...
+  (def $pw (d/db (connect "partner-works")))
+
+  (def bad-hostnames (unknown-hostnames $uw $pw))
+
+  ;; See how many users are in each of these hostnames
+  (def bad-hostname-counts (counts-by-hostname $uw bad-hostnames))
+
+  ;; Then if you want, transfer them to turbovote.org:
+  (def bad-hostname-users (users-by-hostname $uw bad-hostnames))
+  (transfer-hostname (connect "user-works")
```
li-at-dw: does it matter to `(connect "user-works")` here instead of using `uw-conn`?

dev-src/user_works/transfer_hostnames.clj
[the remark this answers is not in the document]
eggsyntax: I _think_ you can? But I didn't try it :)

dev-src/user_works/transfer_hostnames.clj
mrichards42: good question -- it doesn't, but I was intending the examples to be self-contained, and the second example didn't define `uw-conn`

mrichards42: connecting can take some time, but datomic caches connections to the same db, so calling `connect` a bunch of times shouldn't be a problem either.

PR #136, opened 2019-03-13
User ready bug

mrichards42:
[Pivotal card](https://www.pivotaltracker.com/story/show/164600883)

This fixes a subtle bug where the user-ready event could be sent multiple times. It also changes the default sending delay to 5 minutes instead of 1.


mrichards42:
Messages are keyed by their identifiers (or really, a hash of their identifiers), so when you submit a new scheduled message the old one is overwritten.

The idea behind searching for an existing message before scheduling a new one is that we don't want to accidentally schedule a message we should send, either because it was already sent, or because we shouldn't have sent it. If we're just rescheduling, the presence of an existing message is enough to know both of those things.

li-at-dw:
ah, ok.  It's the keying off of `signup` instead of `user-ready` that's the problem.

mrichards42:
> ah, ok. It's the keying off of signup instead of user-ready that's the problem.

exactly -- sorry I wrote a long explanation in the bug card and I thought I gave similar background in this PR, but after a second reading I didn't actually explain that part in this PR

PR #140, opened 2019-06-14
Update to Java 11 and Clojure 1.10.1

gws:
In addition to updating these large bits of software, we can get some quick memory efficiency and configuration wins out of this as well.

The full list of improvements are:

- Java 11
  - You may notice the `yourkit` image was replaced - we haven't been loading that in the Docker entrypoint since the move to Kubernetes, so this shouldn't represent a loss of functionality.
- Clojure 1.10.1
- No longer running as root by default
- Simplify `lein` usage in the Dockerfile
- No longer require `DEPLOYMENT_NAME`
- No longer require `CGROUP_HEAP`
- No longer require `MaxRAMFraction`
- Allow administrators to adjust `JVM_OPTS` and `JVM_MEMORY_OPTS` directly and separately if needed. In almost all cases, the defaults are fine, but when they aren't, you need more power than an on/off switch.

A full explanation of the memory usage improvements can be found in [ADR 2](https://github.com/democracyworks/internal-docs/blob/master/architecture/adr-0002-java-11.md).

It might be easier to see the relationships between the file changes by reviewing each commit in order.

There was a related discussion on [turbovote-admin-http-api#62](https://github.com/democracyworks/turbovote-admin-http-api/pull/62).


project.clj
```diff
@@ -3,28 +3,28 @@
...
                  [com.datomic/datomic-pro "0.9.5703"
                   :exclusions [org.slf4j/slf4j-nop
                                org.slf4j/slf4j-log4j12]]
-                 [com.amazonaws/aws-java-sdk-dynamodb "1.11.448"]
+                 [com.amazonaws/aws-java-sdk-dynamodb "1.11.572"]
                  [ch.qos.logback/logback-classic "1.2.3"]
-                 [democracyworks/kehaar "1.0.0"]
-                 [org.clojure/core.async "0.4.474"]
-                 [prismatic/schema "1.1.9"]
+                 [democracyworks/kehaar "1.0.2"]
```
mrichards42: oh, it occurs to me that if we're going to update deps across all our services, I have an open kehaar PR that just needs to be merged and released

project.clj
```diff
@@ -3,32 +3,33 @@
...
+                 [org.clojure/tools.cli "0.4.2"]
                  [com.rpl/specter "1.1.2"]
-                 [mount "0.1.14"]]
+                 [mount "0.1.16"]]
   :plugins [[com.pupeno/jar-copier "0.4.0"]]
-  :java-agents [[com.newrelic.agent.java/newrelic-agent "4.8.0"]]
+  :java-agents [[com.newrelic.agent.java/newrelic-agent "5.1.1"]]
   :jar-copier {:java-agents true
                :destination "resources/jars"}
+  :jvm-opts ["-Dnewrelic.config.app_name=user-works"]
```
mrichards42: Do we need both this and `NEW_RELIC_APP_NAME` in the configmaps?

gws: Yes*, but if you look at the bottom of the now-removed `newrelic.yml` you'll see that the name appears in fewer places now :joy:

* Strictly, no. But, without it, the New Relic agent dumps a stacktrace and won't start (as opposed to just not finding the license key and not starting), which is pretty ugly and I'm 99% sure will be confusing for people. Check out the stacktrace when you drop this line and run `lein test`. When we run this in production, we're not starting it from `lein`, so we still need to configure it there in the `ConfigMap`.

project.clj
mrichards42: kehaar 1.0.3 is out

gws: Got it!

PR #142, opened 2019-07-24
Update remap chore to send users to queue user-by-user instead of in OCD-ID batch

dzcode:
Paired with @solomonsokoya 
Asana ticket: https://app.asana.com/0/1127682668253243/1127683368758684

This creates two new queues in user-works, one that takes a set of OCD-IDs and returns user IDs, and one that takes a single user ID and remaps it's OCD-IDs. This doesn't yet solve the issue of sending/re-sending notys to users; at the moment, this should only be run for users without upcoming elections. Two questions commented inline.

**Housekeeping**

- [ ] Define config vars in krakenstein:
    - https://github.com/democracyworks/krakenstein/pull/90/files
- [x] Handle `schema` validation errors in calls to user-works
- [x] Remove REMOVEME buildkite commit before merge


dev-src/user_works/remap.clj
```diff
@@ -20,21 +23,67 @@
...
+      (println "Remapped districts for" (:updates results) "users")
+      (println "No updates needed for" (:no-updates results) "users")
+      (when (not-empty errors)
+        (println "Errors remapping" (count errors) "users")
+        (println "******** Errors *******")
+        (pprint errors)))
     (rmq/close rabbit-conn)
-    (println "Done")))
+    (println "Done")
+    (System/exit 0)))
```
dzcode: We modeled this script off of the one here https://github.com/democracyworks/election-authority-works/blob/master/dev-src/election_authority_works/import.clj and included this `System/exit` from that as well. Without this line, the process hangs waiting on something to be closed (though it was unclear what) and then ends ~30 seconds later. The `election-authority-works/import` script linked above has the same problem when this line is removed. If that is indicative of something wrong, happy to investigate further

dev-src/user_works/remap.clj
```diff
@@ -20,21 +23,67 @@
...
+      (= :ok status) user-ids
+      :else (throw (ex-info (str "Error while finding users for ocd-ids:" ocd-ids)
+                            {:msg msg :response response})))))
+
+(defn remap-users
+  [users remap-fn]
+  (let [errors (atom [])
+        updates (atom 0)
+        no-updates (atom 0)]
+    (doall (pmap
```
dzcode: Right now, this seems to send all of the requests to rmq at once; @mrichards42 you had suggested in slack sending in small batches with a rolling queue of 10 user requests out at any given time. Is that necessary to do explicitly, or does `pmap` take care of doing something similar?

resources/config.edn
```diff
@@ -62,6 +62,12 @@
                                 :response true
                                 :prefetch-limit #resource-config/edn #resource-config/env "USER_WORKS_USER_IDS_BY_HOSTNAME_PREFETCH"
                                 :threads #resource-config/edn #resource-config/env "USER_WORKS_USER_IDS_BY_HOSTNAME_THREADS"}
+                               {:queue "user-works.user.user-ids-by-ocd-ids"
+                                :f user-works.handlers/user-ids-by-ocd-ids
+                                :response true
+                                ;; TODO define these in prod and krakenstein
+                                :prefetch-limit #resource-config/edn #resource-config/env "USER_WORKS_USER_IDS_BY_OCD_IDS_PREFETCH"
+                                :threads #resource-config/edn #resource-config/env "USER_WORKS_USER_IDS_BY_OCD_IDS_THREADS"}
```
dzcode: These and the other vars below need to be added to krakenstein before this is merged

dev-src/user_works/remap.clj
```diff
@@ -20,21 +23,67 @@
...
+            (fn [user-id]
+              (try
+                (let [result (async/<!! (remap-fn {:user-id user-id}))]
```
mrichards42: I might split this into a few functions and try to get rid of the atoms.  Something like:

```clj
(defn re-geocode-user [remap-fn user-id]
  (let [response (async/<!! (remap-fn {:user-id user-id}))]
    ;; for logging -- you could also put the log call here, which arguably
    ;; might be simpler. This function is going to be run in a thread, and
    ;; logback (our logging library) is thread-safe, so it should be ok, but
    ;; just be aware of that.
    (assoc response :user-id user-id)))

(defn process-response [result-map response]
  ;; Since we're talking about a lot of users, I'd say we should log the
  ;; responses as we go, instead of adding them to a list of errors
  (when-not (= :ok (:status response))
    (log/error "Error geocoding user " (:user-id response) (pr-str response)))
  (let [k (cond
            (not= :ok (:status response)) :errors
            (:updated-address response) :updates
            :else :no-updates)]
    (-> result-map
        (update :total inc)
        (update k inc))))

(defn remap-users [users remap-fn]
  (->> users
       (pmap (partial re-geocode-user remap-fn))
       (reduce process-response {:total 0, :errors 0, :updates 0, :no-updates 0})))
```

mrichards42: Alternatively, it might actually be more useful to just print the raw response for each user, instead of a summary + errors. So something like:

```clj
(defn remap-users [users remap-fn]
  (->> users
       (pmap (fn [user-id] [user-id (async/<!! (remap-fn {:user-id user-id}))]))
       (run! println)))
```

And then you run the command and `tee` the output to a file.
```

File: src/user_works/handlers.clj [hunk 5]
```diff
@@ -551,6 +556,16 @@
       util/strip-namespaces-from-message
       handle-errors))
 
+(defn-traced update-geocoding-by-user*
```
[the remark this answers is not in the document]
dzcode: That seemed to work, thanks!

resources/config.edn
```diff
@@ -76,7 +82,13 @@
                                 :f user-works.handlers/update-geocoding
                                 :response true
                                 :prefetch-limit #resource-config/edn #resource-config/env "USER_WORKS_OCD_UPDATE_GEOCODING_PREFETCH"
-                                :threads #resource-config/edn #resource-config/env "USER_WORKS_OCD_UPDATE_GEOCODING_THREADS"}]
+                                :threads #resource-config/edn #resource-config/env "USER_WORKS_OCD_UPDATE_GEOCODING_THREADS"}
+                               {:queue "user-works.ocd.update-geocoding-by-user"
```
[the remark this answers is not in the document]
dzcode: Good idea. Leaving this comment open to remind myself that we'll need to remove the relevant env vars from krak + prod as well

dev-src/user_works/remap.clj
gws: If you do end up needing control over concurrency, [`pipeline-blocking`](https://clojuredocs.org/clojure.core.async/pipeline-blocking) is the right thing to use for I/O operations. `pmap` is more useful when applied to workloads for which the parallelism is an implementation detail, e.g. you want it to use a number of threads that's appropriate for the machine on which it's running. See [reborg's comment at the very end of the ClojureDocs `pmap` entry](https://clojuredocs.org/clojure.core/pmap) for how this gets chosen.

dev-src/user_works/remap.clj
```diff
@@ -2,39 +2,81 @@
...
     (init-external-service conn
-                           remap-ocd-ids-chan
-                           "user-works.ocd.update-geocoding")
+                           remap-user-ocd-ids-chan
+                           "user-works.user.update-ocd-ids"
+                           (seconds->milliseconds 60))
+    (init-external-service conn
+                           find-users-by-ocd-ids-chan
+                           "user-works.user.user-ids-by-ocd-ids"
+                           (seconds->milliseconds 60))
```
mrichards42: This works, so you don't need to change this, but fyi calling `kehaar.wire-up/external-service` is an older style. These days it's preferable to use `kehaar.configured` and then configure kehaar and rabbitmq like you would in `resources/config.edn`.

Here's an example of a [config](https://github.com/democracyworks/partner-works/blob/a799f806e11630aac5a6cc949a1480a9a89fce4d/dev-src/partner_works/export_campus_addresses.clj#L10-L27) in a new chore, and this is how you would [set up](https://github.com/democracyworks/partner-works/blob/a799f806e11630aac5a6cc949a1480a9a89fce4d/dev-src/partner_works/export_campus_addresses.clj#L63-L64) and [tear down](https://github.com/democracyworks/partner-works/blob/a799f806e11630aac5a6cc949a1480a9a89fce4d/dev-src/partner_works/export_campus_addresses.clj#L68-L71) all the state.

solomonsokoya:
Hey @mrichards42 would like for you to review this general approach. Still have a few small things to clean up.

dev-src/user_works/remap.clj
```diff
@@ -20,21 +23,67 @@
...
+                           "user-works.ocd.update-geocoding-by-user")
+    (init-external-service conn
+                           find-users-by-ocd-ids-chan
+                           "user-works.user.user-ids-by-ocd-ids")
     conn))
 
+(defn users-by-ocd-ids
+  [ocd-ids]
+  (let [msg {:ocd-ids ocd-ids}
+        {:keys [status user-ids] :as response} (async/<!! (find-users-by-ocd-ids msg))]
```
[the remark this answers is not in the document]
solomonsokoya: How long do you think is enough?

dev-src/user_works/remap.clj
```diff
@@ -2,39 +2,123 @@
...
+(defn remap-user-if-no-elections
+  [remap-fn ocd-ids-with-elections user-id]
+  (let [user (async/<!! (find-user-by-user-id {:id user-id}))
+        user-ocd-divisions (get-in user [:user :addresses :registered :ocd-divisions])]
+    (println "user:" user)
+    ;; Only remap users without upcoming elections
+    (if-not (seq (filter ocd-ids-with-elections user-ocd-divisions))
+      (async/<!! (remap-fn {:user-id user-id}))
+      {:status :has-election
+       :user-id user-id})))
```
mrichards42: I'd be interested in having some rough benchmarks to get an idea of how much overhead it is to check users for elections vs just send them to the queue without checking. If the majority of the time is dominated by calling the civic api (or making sure we're within their rate limits) it doesn't really matter. But if it takes a while to determine which users have elections, it might be worth having an option to run this chore without the check so that it can be done faster.

mrichards42: When we exclude users with elections, we'll probably want to include the matching election ocd-id in the log output, so that we can know which ocd-ids we need to run again once the election is over.

dev-src/user_works/remap.clj
```diff
@@ -2,39 +2,123 @@
...
+  (let [msg {:ocd-ids ocd-ids}
+        {:keys [status user-ids] :as response} (async/<!! (find-users-by-ocd-ids msg))]
+    (cond
+      (= :ok status) user-ids
+      :else (throw (ex-info (str "Error while finding users for ocd-ids:" ocd-ids)
+                            {:msg msg :response response})))))
+
+(defn find-ocd-ids-with-upcoming-elections
+  []
+  (let [upcoming-elections (async/<!! (find-upcoming-elections {}))]
```
mrichards42: Assuming we want to exclude users with a current election, we will want to have more robust error checking on this call. If it fails for any reason, we'll want to halt the script immediately, since otherwise we would treat all users as if they did not have an election.

src/user_works/handlers.clj
```diff
@@ -482,6 +472,21 @@
       util/strip-namespaces-from-message
       handle-errors))
 
+(def user-ids-by-ocd-ids-schema
+  {:ocd-ids [s/Str]})
```
mrichards42: For consistency's sake, we might want to use `:district-divisions` for this key, and `address/division-regex` as the schema (instead of `s/Str`)

mrichards42:
Also heads up that you're going to have some conflicts with #143 once that's merged

gws:
Talking with Mike, It's looking like I won't merge #143 until next week. I don't know what the timeframe is like for this PR but if it makes it easier on you @solomonsokoya I can handle any conflicts on my end after this gets merged.

dev-src/user_works/remap.clj
dzcode: We're still working on other parts of the implementation + testing, but @gws can you take a look at this part and let us know if it's what you had in mind?
https://github.com/democracyworks/user-works/pull/142/files#diff-0b78e433790ba82b4ec564921851cc91R103-R119

gws: @dzcode Yeah, that looks great structurally and does a good job of hiding the use of `core.async` in `remap` :+1:

dev-src/user_works/remap.clj
```diff
@@ -2,39 +2,119 @@
...
+            (or (seq (:added-ocd-ids response)) (seq (:removed-ocd-ids response))) :updates
+            :else :no-updates)]
+    (-> result-map
+        (update :total inc)
+        (update k inc))))
+
+(defn remap
+  [users remap-fn]
+  (let [out-chan (async/chan)]
+    (async/pipeline-blocking 20
```
mrichards42: 👍 I think this was the right way to go, since it gives you more control over how many requests you can have out, and I believe you're going to want to have the timeout on the remap queue be linked to this. Here's my thought process:

TL;DR: it looks like what you have (60 second timeout and 20 concurrents) will work just fine, but I wanted to write all this down and do the math so I had an idea of what might go into that calculation -- and to talk it out in case folks see something missing in my calculations.

We need the timeout for a single message to `user-works.user.update-ocd-ids` to be at least the same as the timeouts we have for external services (civic-info api and datomic). I expect that most of the time will be dominated by the http request to the civic info api (though depending on how fast we make requests, I suppose it's possible that datomic backs up -- but if that happens we'll have bigger fish to fry).

Also, our (new) civic api code does [retries with a backoff](https://github.com/democracyworks/user-works/blob/40309769bd7e892fa10b1afa827fb64d1e26c32c/src/user_works/civic_info.clj#L65-L69) to handle rate-limiting. Our rate limit is 100/s, so there isn't any point in submitting more than 100 messages per second anyways.

We only want to timeout if user-works is unavailable or unresponsive, not b/c we sent messages too fast.
So we need to consider the number of queue consumers in this calculation. Each user-works replica should be able to handle `USER_WORKS_USER_UPDATE_OCD_IDS_THREADS` messages at a time.

So I think this looks like:

```
ideal-timeout = (max-civic-api-timeout * max-civic-api-retries) + datomic-timeout
ideal-concurrency = replicas * USER_WORKS_USER_UPDATE_OCD_IDS_THREADS
```

except we aren't always going to know `number-of-replicas`. we could add extra concurrency if we were to allow for a slightly higher timeout as well, so maybe:
```
concurrency = target-replicas * USER_WORKS_USER_UPDATE_OCD_IDS_THREADS
expected-queued-messages = concurrency - ideal-concurrency
timeout = ideal-timeout * (1 + expected-queued-messages)
```


So I think we're talking about roughly:
```
max-civic-api-timeout = 3s
max-civic-api-retries = 3
datomic-timeout = ?? (I think this defaults to 10s)
ideal-timeout = (3s * 3) + 10s = 19s
# assuming a concurrency of 20, we get
replicas = min: 3 (where we are now); max: a lot -- I think we ran 50ish last fall?
ideal-concurrency = 3 * 6 = 18
concurrency = 20
expected-queued-messages = 20 - 18 = 2
timeout = 19s * (1 + 2) = about a minute
```

src/user_works/entities/user.clj
```diff
@@ -362,3 +377,18 @@
...
+  [user address-id ocd-ids-to-retract ocd-ids-to-add]
+  (let [additions
+        (map
+         (fn [ocd-id]
+           [:db/add address-id :address/ocd-division-id ocd-id]) ocd-ids-to-add)
+        retractions
+        (map
+         (fn [ocd-id]
+           [:db/retract address-id :address/ocd-division-id ocd-id])
+         ocd-ids-to-retract)
```
mrichards42: looks like this is code you copied from `handlers`, but fyi this (and above) is a perfect case for `for` making this more readable:
```clj
(for [ocd-id ocd-ids-to-retract]
  [:db/retract address-id :address/ocd-division-id ocd-id]) 
```

src/user_works/entities/user.clj
```diff
@@ -362,3 +377,18 @@
             db
             date)
        (into {})))
+
+(defn-traced update-registered-address-ocd-ids-db-only
```
mrichards42: This name is pretty long -- do we need `db-only` in the name? All the fns in this namespace should deal with datomic.

Also, this could probably be `update-address-ocd-ids` (or even just `update-ocd-ids`, since they're always on an address), since this fn doesn't do anything anything special with `registered` addresses.

src/user_works/handlers.clj
```diff
@@ -482,6 +468,28 @@
       util/strip-namespaces-from-message
       handle-errors))
 
+(spec/def ::district-division
```
mrichards42: I'd prefer if we didn't mix spec and schema in the same namespace -- we're going to move everything to spec at some point, but I think it would be simpler if we did it in one fell swoop, but I'm open to other opinions if you want to have spec in here.

src/user_works/entities/user.clj
mrichards42: This might also be better suited to `entities.address`, since it only deals w/ addresses (and there are quite similar functions in that namespace, so this might be an opportunity to re-use one of those).

deploy/tv/development/user-works.configmap.yml
```diff
@@ -30,8 +30,10 @@ data:
   USER_WORKS_USER_SIGNUPS_BY_HOSTNAME_THREADS: "1"
   USER_WORKS_REPORTING_QUERY_PREFETCH: "1"
   USER_WORKS_REPORTING_QUERY_THREADS: "1"
-  USER_WORKS_OCD_UPDATE_GEOCODING_PREFETCH: "12"
-  USER_WORKS_OCD_UPDATE_GEOCODING_THREADS: "6"
+  USER_WORKS_USER_IDS_BY_OCD_IDS_PREFETCH: "12"
+  USER_WORKS_USER_IDS_BY_OCD_IDS_THREADS: "6"
```
mrichards42: I don't think we have any other need for a queue that takes a list of ocd-ids to include and a list of ocd-ids to exclude, so this is going to be pretty much exclusive to this chore -- and I think eventually we won't even want this queue, since this is a workaround to deal w/ the way we send election notys, and that's the real problem that should be fixed. Since we're only going to be running this chore one at a time, let's make this 1 thread and 1 prefetch.

dev-src/user_works/remap.clj
```diff
@@ -2,39 +2,119 @@
...
+      :else (throw (ex-info (str "Error while finding users for ocd-ids:" ocd-ids)
+                            {:msg msg :response response})))))
+
+(defn find-ocd-ids-with-upcoming-elections
+  []
+  (let [{:keys [status elections] :as response} (async/<!! (find-upcoming-elections {}))]
+    (when (not= :ok status)
+      (throw (ex-info "Couldn't find upcoming elections" {:response response})))
+    (->> elections
+         (reduce (fn [acc election] (into acc (:district-divisions election))) #{})
```
mrichards42: I believe this line is equivalent to `(mapcat :district-divisions)`

dev-src/user_works/remap.clj
```diff
@@ -2,39 +2,119 @@
...
+         (map :ocd-id)
+         set)))
+
+(defn remap-user-if-no-elections
```
mrichards42: I don't see a check for `if-no-elections` in this fn

test/user_works/entities/user_test.clj
```diff
@@ -233,31 +234,60 @@
...
+  (testing "finds a single user with no upcoming elections"
+    (let [params (user-with-ocd-ids "becky" #{"ocd-division/country:us/state:oh"})
+          {:keys [db-after user-id]} (users/create-user params)
+          found (users/find-uids-by-ocd-ids db-after
+                                            #{"ocd-division/country:us/state:oh"}
+                                            #{"ocd-division/country:us/state:ny"})]
+      (is (= 1 (count found)))
+      (is (= user-id (first found)))))
 
-  (testing "excludes user from other hostname"
```
mrichards42: any reason to remove these tests?

test/user_works/entities/user_test.clj
```diff
@@ -233,31 +234,60 @@
...
+   :last-name "Summers"
+   :hostname "slayer.turbovote.org"
+   :language-preference :es
+   :addresses {:registered {:street "123 Main St"
+                            :city "Sunnyvale"
+                            :state "CA"
+                            :zip "90639"
+                            :ocd-divisions (set ocd-ids)}}})
+
+(deftest find-uids-by-ocd-ids-test
```
mrichards42: Given how much trouble it was to come up w/ that query, let's add some more tests to make sure it's right:

1. tests against the exclude datatype:
    * nil
    * empty set
    * vector of ocd-ids
2. tests with multiple includes and multiple excludes

mrichards42: It might be easiest to transact all the test data once (before any of the `testing` forms) and then run all the queries against that test data

test/user_works/entities/user_test.clj
dzcode: Looks like a mistake, thank you for catching

deploy/tv/development/user-works.configmap.yml
dzcode: Where do the numbers 12 and 6 come from for the rest of these? And should the `UPDATE_GEOCODING` queue be the same, since that's also only used here?

src/user_works/handlers.clj
dzcode: When we used `schema`, errors were getting caught by kehaar on the way back, and so the response looked like it succeeded when in fact it hadn't. I can take another look to see if I can make it catch the errors another way

mrichards42: That's a known issue that will be solved when we move to spec. It's from [this line](https://github.com/democracyworks/user-works/blob/9cd44e2070d7f93b3f522af0dea200d275c1ed0c/src/user_works/handlers.clj#L76) in handle errors, where we just assoc the data on.

We could solve it pretty easily by _not_ associng that data, or by associng stringified data (or doing either only in the case of schema errors.

I agreed that these errors are problematic, but I'd prefer if we fixed the problem everywhere instead of in a single handler.

deploy/tv/development/user-works.configmap.yml
mrichards42: The other numbers are pretty arbitrary, though I know gordon has talked about wanting to do some work on sizing those more appropriately.

I assume you mean `UPDATE_OCD_IDS`? There's a long discussion in [this comment](https://github.com/democracyworks/user-works/pull/142#discussion_r311301032) about how the thread count figures into that one. But that one is different because `IDS_BY_OCD_IDS` is only called once in this script, whereas `UPDATE_OCD_IDS` is called for each user, and we want to be able to process many users concurrently. Also `IDS_BY_OCD_IDS` should go away at some point, so it's a workaround, whereas `UPDATE_OCD_IDS` is the business logic for this chore, and will always exist in some form.

gws: There's no reasoning behind `12` and `6` as far as I am aware - just hop on that bandwagon, leave them where they're at (unless you know you need more or less), and we'll get to them later this year.

I also loathe knobs like these that could just be reasonable defaults and changed only if necessary...

dev-src/user_works/remap.clj
mrichards42: I thought about this a little bit more, and realized I didn't take `USER_WORKS_USER_UPDATE_OCD_IDS_PREFETCH` into account.

[Prefetch](https://www.rabbitmq.com/confirms.html#channel-qos-prefetch) exists to reduce network chatter and increase throughput. The way it works is (roughly) by delivering up to `prefetch-count` messages to the consumer _before_ the consumer has picked any of them up. These are `unacked` messages -- delivered but not acknowledged. You can think of this as a per-consumer mini-queue, so we have the same concerns as with `expected-queued-messages` above.

The way `kehaar` is structured, each replica of a service counts as a consumer, and each consumer has threads working on the consumer's prefetch mini-queue.


I believe what this means for the worst case scenario, which is what we should base timeout on, is that a message could be sitting in the prefetch mini-queue for an additional `prefetch / threads` messages, so add that to `expected-queued-messages`:

```
expected-queued-messages = concurrency - ideal-concurrency + (prefetch / threads)
# so in our case
expected-queued-messages = 20 - 18 + (12 / 6) = 4
timeout = 19s * (1 + 4) = ~100s
```

Or we could knock prefetch down to 6 for instance (prefetch = threads is also a common pattern in some of our services), and end up w/
```
expected-queued-messages = 20 - 18 + (6 / 6) = 3
timeout = 19s * (1 + 3) = ~80s
```

dev-src/user_works/remap.clj
mrichards42: With the way you've changed the query for ocd-ids this shouldn't be a problem any more.

deploy/tv/development/user-works.configmap.yml
mrichards42: > I also loathe knobs like these that could just be reasonable defaults and changed only if necessary...

Absolutely. Might be worth having kehaar handle `nil`s for these instead of just [missing keys](https://github.com/democracyworks/kehaar/blob/86c704432e4e8cc8b37162ab43014dbbb20dfa7f/src/kehaar/configured.clj#L140-L145).

dev-src/user_works/remap.clj
mrichards42: But also again, since I'm talking about the worst-case scenario w/ this timeout, a lot of things would have to go wrong to need the 80 or 100s timeout, so you can still leave these numbers where you have them already.

src/user_works/handlers.clj
```diff
@@ -225,6 +163,48 @@
   [user]
   (reduce geocode-user-address user (:addresses user)))
 
+(defn update-geocode-registered-address
```
dzcode: Remove logging in this function

src/user_works/handlers.clj
```diff
@@ -225,6 +163,48 @@
...
+        user-id (:user/id user)
+        current-ocd-divisions (:address/ocd-division-id registered-address)
+        {updated-address :address geocode-status :status}
+        (geocode-address* (util/denamespace-map
+                           registered-address
+                           address/address-keys))
+        updated-ocd-divisions (:ocd-divisions updated-address)]
+    (if (not= geocode-status :ok)
+      ;; Log and send back the non-updated address, since we aren't changing it
+      (do (log-geocoding-result user-id geocode-status registered-address)
```
dzcode: move this outside of `if`

src/user_works/handlers.clj
```diff
@@ -225,6 +163,48 @@
...
+          {:status geocode-status
+           :user-id user-id
+           :address registered-address})
+      (let [ocd-ids-to-retract (set/difference current-ocd-divisions updated-ocd-divisions)
+            ocd-ids-to-add (set/difference updated-ocd-divisions current-ocd-divisions)]
+        (if (= updated-ocd-divisions current-ocd-divisions)
+          (log/infof "No ocd-id updates for user: %s" user-id)
```
dzcode: remove this log, `when not=` above

src/user_works/handlers.clj
```diff
@@ -225,6 +163,48 @@
...
+          (do (log/infof "Re-mapping ocd-ids for user %s, address: %s" user-id registered-address)
```
dzcode: do all in one line

dev-src/user_works/remap.clj
```diff
@@ -2,39 +2,122 @@
...
+      :else (throw (ex-info (str "Error while finding users for ocd-ids:" include-ocd-ids)
+                            {:msg msg :response response})))))
+
+(defn election-ocd-ids
+  []
+  (let [{:keys [status elections] :as response} (async/<!! (upcoming-elections {}))]
+    (when (not= :ok status)
+      (throw (ex-info "Couldn't find upcoming elections" {:response response})))
+    (->> elections
+         (mapcat #(:district-divisions %))
```
dzcode: don't need anon fn

dzcode:
Completed test plan: https://docs.google.com/document/d/1cnNX1jZoLGMG5zpzeWBQuB2heNgztQlX6VBtmfG4NFE/edit#

PR #143, opened 2019-08-02
Remove address-works

gws:
# Context

As we have changed our usage of `address-works` over time, we are now at a point where we make a call to that service solely to fetch OCD IDs from the Google Civic Info API.

Given that this call currently occurs in the "hot path" of a user creation request, we can get the following benefits by calling the Civic Info API directly:

- Reduce the load on RabbitMQ
- Save resources by not deploying `address-works`
- Optimize the request to the Google Civic Info API to ask for only the fields necessary to obtain OCD Division IDs
- Have better control over the request concurrency, since it is now governed by the number of threads handling the requests that, in turn, perform the requests to the Google Civic Info API. Before, we would have to right-size `address-works` as well to ensure we weren't making `user-works` wait around for `address-works` threads to become available, and getting that right between the two services is tricky and not obvious. Now, it's 1-1 with the number of threads we have processing user creates/updates.

# Extras

- New Relic is removed
- OpenTracing has been added
- Miscellaneous cleanup (Buildkite, unnecessary project dependencies, etc.)
- Updated documentation

# References

1. [Asana task](https://app.asana.com/0/1126811134707123/1126811134707146/f)


src/user_works/civic_info.clj
```diff
@@ -0,0 +1,142 @@
...
+                          :ocd-divisions (extract-ocd-ids (get response :body))}
+                         (if (= 404 (get response :status))
+                           {:status :not-found}
+                           {:status :error})))]
+    (try
+      (response-map
+       (diehard/with-retry {:policy *retry-policy*}
+         (representative-info-by-address client
+                                         {:address address-str
+                                          :fields "divisions"})))
```
gws: This is a neat optimization: Google APIs allow for [requesting only specific fields](https://developers.google.com/civic-information/docs/v2/parameters) for performance reasons. You could write a different function taking advantage of the protocol but passing in a different set of fields (like if you wanted to go back to getting standardized addresses from Google).

gws:
I'm still working out a couple kinks after I made some changes to put some of the `:not-found` and `:error` logic back to the way it was before.

src/user_works/handlers.clj
```diff
@@ -113,28 +95,11 @@
...
-    (cond
-      (nil? response)
-      {:status :timeout}
-
-      (empty? ocd-divisions)
-      {:status :not-found}
-
-      :else
-      {:ocd-divisions ocd-divisions
-       :status :ok})))
```
gws: This wasn't really adding anything since we have to re-test the return values when we do similar logic in `update-geocode-registered-address`. In future if we _do_ need to refactor this code, I think we should find a way to reuse `geocode-address` instead of duplicating the `cond` logic here.

src/user_works/handlers.clj
```diff
@@ -159,22 +124,34 @@
...
-      :ok
+        {:keys [ocd-divisions] :as response}
+        (civic-info/get-ocd-division-ids civic-info-client
+                                         registered-address-map)]
+    (cond
+      (nil? ocd-divisions)
+      (log/errorf "Google Civic Info API response was: %s" response)
+
+      (empty? ocd-divisions)
+      (log/infof "No OCD IDs to update for address: %s" registered-address)
```
gws: These two error conditions were tested below (more subtle), but this ordering looks more like other logic in this namespace and probably easier to follow.

src/user_works/civic_info.clj
```diff
@@ -0,0 +1,142 @@
...
+(defn- extract-ocd-ids
+  "Extracts OCD IDs from the top-level keys of the \"divisions\" entry in the
+  Google Civic Info API response as well as from any \"alsoKnownAs\" entries
+  that may exist.
+
+  Returns a set of OCD IDs."
+  [body]
+  (set (specter/select
+        (specter/multi-path MAP-KEYS [MAP-VALS "alsoKnownAs" ALL])
+        (get body "divisions"))))
```
gws: Replaces calls to these two functions: https://github.com/democracyworks/address-works/blob/a8860f4492836919797999cf31beda38c0f292a6/src/address_works/results/google_civic_info.clj#L17-L35

resources/config.edn
```diff
@@ -103,4 +99,6 @@
                                 :timeout 30000}]}}
  :datomic
  {:uri #resource-config/env "USER_WORKS_DATOMIC_URI"
-  :partition :user-works}}
+  :partition :user-works}
+ :civic-info
+ {:api-key #resource-config/env "GOOGLE_CIVIC_API_AUTH_KEY"}}
```
mrichards42: Not seeing this in the user-works configmaps -- is this in one of the shared configmaps?

src/user_works/civic_info.clj
mrichards42: So, knowing (almost) nothing about specter, is there a simple explanation for why
```clj
(select (multi-path MAP-KEYS [MAP-VALS "alsoKnownAs" ALL]) ...)
```
returns a flattened seq instead of a list of vectors?
```clj
([division1 division2] [also-known-as-1])
```

src/user_works/civic_info.clj
```diff
@@ -0,0 +1,142 @@
...
+  #{"concurrentLimitExceeded" "rateLimitExceeded" "servingLimitExceeded"
+    "userRateLimitExceeded"})
+
+(defrecord Client [api-key]
+  RepresentativesAPI
+  (representative-info-by-address
+    [client params]
+    {:pre [(contains? params :address)]}
+    (tracing/with-span [s {:name (str `representative-info-by-address)}]
+      (let [{:keys [api-key]} client
```
mrichards42: I think `api-key` is already available since this is inside a `defrecord` method

src/user_works/civic_info.clj
```diff
@@ -0,0 +1,142 @@
...
+            client-params {:accept :json
+                           :as :json-string-keys
+                           :coerce :always
+                           :connection-timeout 3000
+                           :query-params (merge {:key api-key} params)
+                           :socket-timeout 3000
+                           :throw-exceptions? false}]
+        (client/get representatives-base-url client-params)))))
```
mrichards42: It took me a second to figure out which `client` we're talking about (http or civic-info). I'm not sure what the best way is to deal w/ this overloaded name. I known `clj-http`'s `client` is used all over the place and is standard, but since we also have a `Client` record it's a little confusing at first.

test/user_works/test_helpers.clj
```diff
@@ -1,15 +1,57 @@
...
+
 (defn with-fresh-db
   [tests]
   (let [uri (str "datomic:mem://user-works-" (UUID/randomUUID))]
     (-> (mount/except #{#'queue/rabbit-resources})
         (mount/swap {#'entities/datomic-config
-                     {:uri uri, :partition :user-works}})
+                     {:uri uri, :partition :user-works}
+                     #'handlers/civic-info-client
+                     (mock-representatives-api representatives-api-200-response)})
```
mrichards42: It doesn't seem like a mock civic-info client belongs in a function named `with-fresh-db`. Would it make sense to have a second fixture fn `with-mock-civic-info` or something like that?

gws: Yes, definitely!

src/user_works/civic_info.clj
gws: I agree, a rename is in order for the Civic Info implementation.

resources/config.edn
gws: Good eye, but yes, it should already be available to `user-works` through a shared `ConfigMap`.

src/user_works/civic_info.clj
gws: Yeah, the very short explanation is you have to pay attention to what the navigators are supposed to return.

[Some navigators](https://github.com/nathanmarz/specter/wiki/List-of-Navigators#filterer) are defined to return vectors, and those are called out in the docs. You can [see the difference in the first two multi-path examples](https://github.com/nathanmarz/specter/wiki/List-of-Navigators#multi-path).

src/user_works/civic_info.clj
gws: Thanks, this is a bad Clojure habit I picked up from using Component way back when that I need to kick.

src/user_works/civic_info.clj
mrichards42: thanks!

gws:
Ran a couple signups on staging, got the OCD IDs I was after, all looks good as far as I can see. Going to `master` and then I'll watch some production signups.

PR #144, opened 2019-09-18
Initial NVRD 2019 (and beyond) settings

gws:
The two main changes:

# Raise thread counts for HTTP API handlers

The handlers with raised counts directly handle TurboVote HTTP API requests. The numbers are chosen somewhat arbitrarily, but with a ballpark in mind. These handlers emit RabbitMQ events, and we estimate that a single RabbitMQ producer can achieve at or a bit above 100 messages per second in production due to the average latency present there. Given that most requests should take no more than 500ms to process, 50 threads producing messages every 500ms should be able to saturate the RabbitMQ producer. Therefore we do not want to go much beyond 50 threads per user-works instance - at that point it is better to allow other instances (RabbitMQ producers) to handle requests. Further, we expect to handle most requests faster than the 500ms high-end estimate.

The number 50 also matches the default Jetty request handling thread pool size.

The 25 message prefetch count is more arbitrary - a high prefetch count in our case can risk more data loss if the service is restarted under load (reminder that Kehaar acknowledges all received messages before the service sees them), but set too low and the handlers may not be fully saturated. It has been historically set in other services to half of the thread pool size, so we maintain that here.

# Replace async/put! with async/>!!

async/put! is dangerous to use by itself if backpressure is not respected. We hit conditions during 2018 that resulted in exceptions being thrown due to filling up the implicit 1024-message internal channel buffer.

Users may want to be aware of a buffer (currently 1000 messages) that Kehaar uses internally (as an implementation detail, subject to change) when reading from the channel being put to, so there will be some absorption of spikes, but backpressure will be applied as expected once that fills up.


gws:
@eggsyntax Yes, it's very possible these will change later this week, but also I'd like to establish a good default thread/prefetch count for services that we can use year round, so we may scale out _some_ for NVRD but not _a lot_ (think 8-10 replicas instead of 30-50), and I'm preferring changes that don't need to be un-done post-NVRD (so they don't result in outsized capacity consumption, etc.)

deploy/tv/production/user-works.configmap.yml
```diff
@@ -5,16 +5,16 @@ metadata:
...
-  USER_WORKS_USER_CREATE_PREFETCH: "12"
-  USER_WORKS_USER_CREATE_THREADS: "6"
-  USER_WORKS_USER_DELETE_PREFETCH: "12"
-  USER_WORKS_USER_DELETE_THREADS: "6"
-  USER_WORKS_USER_UPDATE_PREFETCH: "12"
-  USER_WORKS_USER_UPDATE_THREADS: "6"
+  USER_WORKS_USER_CREATE_PREFETCH: "25"
+  USER_WORKS_USER_CREATE_THREADS: "50"
+  USER_WORKS_USER_DELETE_PREFETCH: "25"
+  USER_WORKS_USER_DELETE_THREADS: "50"
```
dzcode: I don't know if it makes a difference, but I think this is only used by the admin console, so wouldn't need to be scaled up - I think @leilakass has been looking at this endpoint so may know better

gws: Ah, you're probably right, I was going by what the HTTP API *can* hit directly. Note that these are maximum values - threads are not allocated until they are needed, and they are pulled from a cached thread pool that `core.async` provides, so nothing will actually be allocated until / unless it's needed.

PR #145, opened 2019-09-19
Stop searching for scheduled user.ready events

gws:
This avoids an entire wasteful roundtrip to scheduled-message-works to accomplish something that is done when the message is scheduled anyway.

Since user updates are one of the highest-volume interactions in our system during signup, cutting the RMQ traffic to 1/3 of what it was for this operation seems like a big win.

**Please scrutinize this carefully!**


mrichards42:
This seems to be the pattern that's used for rescheduling messages -- effectively `reschedule-if-exists` (i.e. update semantics instead of upsert). I know you've talked about making that concept a first-class scheduled-message-works citizen, which I'd definitely be interested in. I think we could use it in election-notification-works as well.

The way I read this, we'd change `user.ready` from being send-once (aside from the inherent race condition) to send-at-least-once, which is probably OK, given that crm-works is the only consumer, but I actually had a conversation today w/ esha+carmen about revamping `user.ready` so that it could be used in more places, so I'm hesitant to change the behavior of this event to "send at least once".

gws:
@mrichards42 This doesn't alter the existing behavior in any way that I can tell - the test that's being done in the removed function will reschedule the message _when one already exists_ - it's already sending `scheduled-message-works` the same message multiple times (just check the logs!)

If you wanted it to be send-once, you'd try to check for an _empty_ set of messages and only send in that case. But that's still subject to the same race where a message can be sent (or deleted in the old system, same semantics though) and then a new user.ready event would be scheduled in its place.

gws:
Oh and the other obvious race that two instances check for a message at the same time, and schedule `user.ready` notifications - the classic "search and then send" pattern that doesn't work without some kind of isolation.

gws:
It's worth mentioning that the previous version of scheduled-message-works did the same thing with this message - it deleted the old version first (matching the identifiers exactly) and then re-inserted it. While I'm not doing exactly the same thing in the current version I'm updating everything about the message that could feasibly change, so the only subtle difference is the message ID doesn't change.

gws:
@mrichards42 **Sorry, nevermind, I totally misunderstood you!** I thought you were talking about the interaction before the user.ready event gets sent. We're still talking about the same race if/when the user.ready event does get sent out by `scheduled-message-works`.

mrichards42:
Oh yeah, sorry that was not the clearest comment. I think we're on the same page, but just to make sure (and for future readers), I meant that the idea is "schedule (and overwrite) the message as long as one is already scheduled", which is also to say "don't schedule the message if it was already sent".

And the race I was talking about is that an existing message could be sent between searching for an existing message and rescheduling it.

PR #147, opened 2020-01-22
Reduce TurboVote memory usage to 4GiB

gws:
I can't think of a good reason that user-works needs to use this much RAM per instance in the TurboVote cluster. I do believe that it was set to 12GiB at one point due to the way tvadmin was performing queries requiring a very large dataset to reside in memory at once, so I've left that deployment alone for this PR.

I've been running load tests with less RAM and everything seems to be fine, but those don't exercise all the handlers, just the those we use at signup time.


dzcode:
Is there any way to check what the RAM usage was during the election imports yesterday? That's the only thing I can think of that could maybe need this much RAM. Otherwise, makes sense to me if your load tests are looking 💯

gws:
@dzcode Do we use the TurboVote admin cluster for imports?

mrichards42:
Chores are done on the tv cluster as a rule (except for admin-specific chores, like user exports). IIRC we did some election imports on the admin cluster around NVRD last year, but in general we shouldn't be doing that since the admin cluster isn't guaranteed to have all the required services for that chore.

mrichards42:
But that's a decent high water mark -- election imports are probably the heaviest single query we do regularly on the tv cluster, since it's "fetch all user ids in an ocd-id" + "stream back the full user record". Just the initial query for user-ids is probably similar to doing a export on a large partner, though obviously not as bad as doing an export of all users.

gws:
I'll hold off on this pending further testing. As noted in Slack, I am suspect of https://github.com/democracyworks/kehaar/blob/master/src/kehaar/core.clj#L213-L216 and we need to investigate whether we're unnecessarily holding onto the heads of our sequences.

gws:
@eggsyntax Yes, we could do this with two deployments, each with the appropriate number of replicas and requests/limits set. Unfortunately there's no way we can ensure the 4GiB replicas receive the streaming search request, but with enough elections being imported we're bound to see it get picked up.

PR #148, opened 2020-01-23
VE-131 Reschedule user ready

mrichards42:
Using democracyworks/scheduled-message-works#92 to reschedule this event asynchronously. I used a light hand w/ this code and avoided much-needed test refactoring so we can get this out the door. I don't think this is any _worse_ than what was there before at least.


mrichards42:
My signup tests in the testing environment passed, but since I'm out tomorrow I'll hold off on merging these until I'm back on Monday so I'm around in case anything goes wrong.

PR #151, opened 2020-02-17
Add prometheus metrics

sduckett:


project.clj
```diff
@@ -4,27 +4,28 @@
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
   :dependencies [[org.clojure/clojure "1.10.1"]
-                 [org.clojure/tools.logging "0.4.1"]
+                 [org.clojure/tools.logging "0.6.0"]
                  [democracyworks/resource-config "1.0.0"]
                  [democracyworks/datomic-toolbox "2.0.5"
                   :exclusions [com.datomic/datomic-pro]]
                  [com.datomic/datomic-pro "0.9.5703"
```
gws: I don't think `ancient` does this by default, but it'd be great to start moving our `com.datomic/datomic-pro` dependencies up to the latest version. I think `0.9.6045` is the latest? I recognize that's out of scope for this PR, though. If there's not a TV Internals ticket for doing this, I may create one today.

sduckett: No, `ancient` doesn't do this; thanks for pointing it out. I'd probably consider it in-scope since I'm updating dependencies as a matter of practice.

sduckett: Covered in https://github.com/democracyworks/user-works/pull/152

PR #152, opened 2020-02-18
Find user-ids by hostnames & ocd-ids

sduckett:
### Overview

This PR adjusts the query behind the channel that `tv-admin-works` pulls user IDs from when doing an export. It supports doing user exports with both hostname and ocd-id filters.

### Performance

When we're only applying one filter, things take one kind of time.

```clojure
user-works.entities.user> (let [ocd-ids #{"ocd-division/country:us/state:ca"}
                                hostnames :all
                                start nil
                                end nil
                                query (uids-with-signup-by-hostnames-and-ocd-id-query
                                       db hostnames ocd-ids start end)]
                            (time (->  query
                                       d/query
                                       count)))

"Elapsed time: 37812.661545 msecs"
658052
```

Then, when we add another second filter, things take another kind of time.
 
```clojure
user-works.entities.user> (let [ocd-ids #{"ocd-division/country:us/state:ca"}
                                hostnames ["moderate-hostname.turbovote.org" 
                                           "big-hostname.turbovote.org"]
                                start nil
                                end nil
                                query (uids-with-signup-by-hostnames-and-ocd-id-query
                                       db hostnames ocd-ids start end)]
                            (time (->  query
                                       d/query
                                       count)))

"Elapsed time: 97281.413736 msecs"
407350
```

[TVF-33](https://democracyworks.atlassian.net/browse/TVF-33)


project.clj
```diff
@@ -4,27 +4,27 @@
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
   :dependencies [[org.clojure/clojure "1.10.1"]
-                 [org.clojure/tools.logging "0.4.1"]
+                 [org.clojure/tools.logging "0.6.0"]
                  [democracyworks/resource-config "1.0.0"]
                  [democracyworks/datomic-toolbox "2.0.5"
                   :exclusions [com.datomic/datomic-pro]]
-                 [com.datomic/datomic-pro "0.9.5703"
+                 [com.datomic/datomic-pro "0.9.6045"
```
gws: :man_cartwheeling:

src/user_works/entities/user.clj
```diff
@@ -117,31 +117,53 @@
...
+   For use with `d/query`.
+
+   Optional filters:
+    - `hostnames` may be a collection of partner hostnames (strings) or `:all`
+    - `ocd-ids` may be a collection of district divisions, or nil.
+
+   Optional bounds:
+   - `start-date`: signups happened after this.
+   - `end-date`: signups happened before this."
+  [db hostnames ocd-ids start-date end-date]
```
gws: Wonder if we're already breaking the signature if it would make sense to accept a `filters` (or similarly-named) argument at this point.

gws: Could potentially simplify the function name, too.

sduckett: Sure, and we've got a few other functions to find by ocd-id (it has exclusions as well!) that could be consolidated, I suppose.

src/user_works/entities/address.clj
```diff
@@ -90,8 +89,7 @@
 
 (defn update-addresses-tx-data
   [user-entity updated-addresses]
-  (let [user-db-id (:db/id user-entity)
-        current-types (->> user-entity
+  (let [current-types (->> user-entity
```
dzcode: nice catch that `user-db-id` was unused

src/user_works/handlers.clj
```diff
@@ -447,15 +444,16 @@
       util/strip-namespaces-from-message
       handle-errors))
 
-(def user-ids-by-hostnames-schema
+(def user-ids-by-hostnames-and-ocd-ids-schema
   {:hostnames (s/cond-pre (s/enum :all) #{s/Str})
+   (s/optional-key :ocd-ids) (s/cond-pre (s/enum :all) #{s/Str})
```
dzcode: I don't think you need the `:all` here

test/user_works/entities/user_test.clj
```diff
@@ -519,43 +508,105 @@
...
-                     db-after :all nil nil)))))
+                     db-after :all #{"ocd-division/country:us/state:nm"} nil nil))))
+      (is (= 7
+             (count (users/find-uids-with-signup-by-hostnames
+                     db-after :all #{"ocd-division/country:us/state:ca"} nil nil))))
+      (is (= 14
+             (count (users/find-uids-with-signup-by-hostnames
+                     db-after :all #{"ocd-division/country:us/state:ca"
+                                     "ocd-division/country:us/state:nm"}
+                     nil nil)))))
```
dzcode: ⭐️

src/user_works/handlers.clj
sduckett: This should look better now.

src/user_works/handlers.clj
```diff
@@ -444,16 +444,16 @@
       util/strip-namespaces-from-message
       handle-errors))
 
-(def user-ids-by-hostnames-and-ocd-ids-schema
+(def user-ids-by-filters-schema
   {:hostnames (s/cond-pre (s/enum :all) #{s/Str})
-   (s/optional-key :ocd-ids) (s/cond-pre (s/enum :all) #{s/Str})
+   (s/optional-key :ocd-ids) (s/cond-pre #{address/division-regex})
```
dzcode: nice change!

test/user_works/entities/user_test.clj
```diff
@@ -531,82 +531,118 @@
         users [user1 user2 user1 user2 user1 user2 user2
                user3 user4 user3 user4 user4 user4 user4]
         db-after (:db-after (last (map users/create-user users)))
-        sorted-suts (->> (users/find-uids-with-signup-by-hostnames db-after :all nil nil nil)
+        sorted-suts (->> (users/find-uids-by-filters db-after
```
dzcode: I know this was there before you, but if you  know... what does `sorted-suts` mean?

test/user_works/entities/user_test.clj
```diff
@@ -531,82 +531,118 @@
...
+                               :start-date nil
+                               :end-date nil})))))
 
     (testing "start-date only"
       (is (= 1
-             (count (users/find-uids-with-signup-by-hostnames
-                     db-after ["paprika.turbovote.org"] nil (nth sorted-suts 3) nil))))
+             (count (users/find-uids-by-filters
+                     db-after {:hostnames ["paprika.turbovote.org"]
+                               :ocd-ids nil
```
dzcode: I think for the nils, I would remove those keys in all but maybe one test (to make sure it's okay with nil given)

src/user_works/entities/user.clj
```diff
@@ -128,7 +128,8 @@
    Optional bounds:
    - `start-date`: signups happened after this.
    - `end-date`: signups happened before this."
-  [db hostnames ocd-ids start-date end-date]
+  [db {:keys [hostnames ocd-ids start-date end-date]
+       :as filters}]
   (let [use-hn (not= hostnames :all)
```
dzcode: Would it be worth explicitly checking that hostnames isn't nil?

dzcode: I assume that would break the query, but don't know what would happen for sure

test/user_works/entities/user_test.clj
sduckett: yeah, I've been thinking about that: probably _sorted sign-up timestamps_.
so, I'm thinking _signup-timeline_ perhaps.

gws: I'm going with *subject-under-tests* as a guess! :game_die:

dzcode: 😮  the more you know 🌈

sduckett: I'm basing my guess on how [they're used as filters](https://github.com/democracyworks/user-works/blob/c3ea3a3386f3608b74ceb5e624c2c9d387a13342/test/user_works/entities/user_test.clj#L637-L642):
```clojure
{:start-date (nth sorted-suts 1)
 :end-date (nth sorted-suts 4)}
```

src/user_works/entities/user.clj
sduckett: I think `nil` and `:all` express the same semantics: Don't filter my results by hostname. We can probably make this more clear.

dzcode: I agree, I think they do - but I don't know what happens if you were to pass `nil` expecting the behavior of all (especially since `nil` `ocd-ids` has that behavior)

sduckett: What you'd run into when passing `nil` and expecting `:all` is an exception about a missing data source in the query.

dzcode: Maybe, though I think it's a way of data firewalling, which is probably good (you have to ask for `:all` specifically, so it'd be harder for a nil-passing bug to lead to a single partner getting data for users across all partners). Regardless, I think it's fine to leave as is given the work anyway

sduckett:
@dzcode / @gws Have I addressed your concerns? If so, I'll merge this.

docker-compose.yml
```diff
@@ -39,7 +39,7 @@ services:
       USER_WORKS_OCD_UPDATE_GEOCODING_THREADS: 2
 
   datomic:
-    image: quay.io/democracyworks/datomic-tx:0.9.5783
+    image: quay.io/democracyworks/docker-pro-transactor:0.9.6045
     environment:
       ALT_HOST: datomic
```
gws: You should be able to drop the `environment` key, and add a `hostname: datomic` entry here. This image will use that hostname as its advertised hostname and therefore be accessible in the docker-compose network. Without the `hostname` entry the hostname inside the container is the container ID, which isn't how other containers will be able to contact it. The default `ALT_HOST` is already localhost (and the new env var is `DATOMIC_ALT_HOST` anyway), so were somebody want to expose this to their local machine they would just be able to expose the ports and things would work fine.

gws: See Krakenstein for inspiration (and DATOMIC_ALT_HOST is actually redundant there as well and could be removed, I think it was me that left that in).

docker-compose.yml
```diff
@@ -39,9 +39,8 @@ services:
       USER_WORKS_OCD_UPDATE_GEOCODING_THREADS: 2
 
   datomic:
-    image: quay.io/democracyworks/datomic-tx:0.9.5783
-    environment:
-      ALT_HOST: datomic
+    image: quay.io/democracyworks/docker-pro-transactor:0.9.6045
```
gws: Oh, I missed this before the merge, the name of this image is `datomic-pro-transactor` - I'll fix this directly on `master`.

PR #153, opened 2020-03-17
Increase user.page threads in admin

mrichards42:
I'm not sure why this was at 1 thread per pod before, but from what I can tell this is a limiting factor in export speed (especially when trying to run more than 1 export at a time).


mrichards42:
Oh, nice digging -- I had a hard time tracing the files back once they changed names too much :) It seems like we've been going with prefetch x2, so I don't mind doing that.

gws:
:+1: to the thread increase!

I don't understand the logic behind prefetch > threads though, I'd suggest keeping prefetch <= threads without a good reason to do it differently - if you do want to keep prefetch=20 threads=10 then keep an eye on the nack/reject rate via RabbitMQ during an export to see if it's effective at accomplishing your goal or if it's just noisy.

mrichards42:
> I don't understand the logic behind prefetch > threads though, I'd suggest keeping prefetch <= threads without a good reason to do it differently

That works for me -- this is probably an expensive enough queue that additional prefetch probably can't be justified anyways.

PR #154, opened 2020-03-25
Add Google Civic Info API latency metrics

gws:
The purpose of this PR is to measure Google Civic Info API latency, using a Prometheus summary metric type. This metric type tracks both a value (in this case, latency in seconds) as well as the number of observations.

I'm aiming to provide a relatively minimal template that is easy for other services to copy to start tracking their own metrics rather than adding all useful metrics. A good followup PR would be to add the metrics from the wishlist in #151.

In addition, I fixed up the `docker-compose` situation a bit so that it runs correctly, assuming a `profiles.clj`.

Inspired by work done in #151.


deploy/tv/user-works.deployment.yml
```diff
@@ -16,6 +16,19 @@ spec:
...
+        ad.datadoghq.com/user-works.check_names: |
+          ["openmetrics"]
+        ad.datadoghq.com/user-works.init_configs: |
+          [{}]
+        ad.datadoghq.com/user-works.instances: |
+          [
+            {
+              "prometheus_url": "http://%%host%%:9090/metrics",
+              "namespace": "user_works",
+              "metrics": ["civic_info_api_duration_s"]
```
sduckett: Do we get the JVM metrics for free, or do we need to declare them here?

deploy/tvadmin/user-works.deployment.yml
```diff
@@ -16,6 +16,19 @@ spec:
...
+        ad.datadoghq.com/user-works.check_names: |
+          ["openmetrics"]
+        ad.datadoghq.com/user-works.init_configs: |
+          [{}]
+        ad.datadoghq.com/user-works.instances: |
+          [
+            {
+              "prometheus_url": "http://%%host%%:9090/metrics",
+              "namespace": "user_works",
+              "metrics": ["civic_info_api_duration_s"]
```
sduckett: repeat: JVM metrics question?

deploy/tv/user-works.deployment.yml
gws: They're being collected by the Prometheus client, but not submitted to Datadog based on this configration - we need to delcare them here as you mentioned. We can pick and choose which ones we submit - the reason I didn't list them all immediately is that they count as custom metrics, so we just need to determine which ones are useful and submit those to avoid extra unwanted Datadog cost.

I don't immediately see a list of the exported metrics in the Java Prometheus client (though I can read some of the hotspot code to see them), but I can add one or two here to get us started if that'd help!

sduckett: There's a fair bit of digging, it seems, but they're called things like `jvm_gc_collection_seconds_count{gc="PS1"}` or `jvm_memory_bytes_used{area="heap"}`.

See also http://prometheus.github.io/client_java/io/prometheus/client/hotspot/package-summary.html

mrichards42: Ah, I didn't realize these were going to count as custom metrics. Would it be worth using the datadog [agent](https://docs.datadoghq.com/integrations/java/?tab=host#overview) to pick up jvm metrics? I'm honestly not sure what this would entail, but by my reading they don't count as custom metrics.

gws: @mrichards42 I think we can, it would just involve a change to run the Java-based Datadog agent instead of the smaller one we run now. I haven't played around with it much but we'd also need to expose the JMX port in each container if that's not open by default.

How about this: I'll take the JVM metrics out of here for now and I'll play around with the Datadog agent in a cluster I have better control of, and that way we can make specific recommendations to Fairwinds.

gws:
Thank you both! Getting this in here as-is and am noting the JVM/JMX followup.

PR #155, opened 2020-03-30
Tvf-32-conflicting-ocd

leilakass:
This addresses [TVF-32](https://democracyworks.atlassian.net/browse/TVF-32?atlOrigin=eyJpIjoiZGNkMDkyYmI2MjE1NDc5ZTlmN2E2Mjk2OTZjYzUwMDciLCJwIjoiaiJ9)

It includes:

1. We will check the Civic Info API response for conflicting ocd-ids before ocd-ids are transacted into the database in user-works. Any users with bad ocd-ids will be assigned the “default” set of ids, consisting of just state and country ids. (bad ocd-ids = ocd-ids where the state doesn’t match the address, invalid pairs of places that match a static list)
2. Maintain a static list of invalid pairs of place to check ocd-ids against

You can see ADR for this work here: https://democracyworks.atlassian.net/wiki/spaces/TV/pages/146243877/ADR+1+Handling+Conflicting+Ocd-Ids.

I will add a few more invalid pairs to the list + a test to cover them once I get clarity from Product but I didn't want to block the significant code from being reviewed


src/user_works/handlers.clj
```diff
@@ -86,13 +86,14 @@
       :not-found (log/warn "Address not found attempting to geocode for"
                            user-id-str ":"
                            (pr-str address))
+      :conflicting (log/warn "Conflicting ocd-ids found attempting to geocode for"
+                             user-id-str ":"
+                             (pr-str address))
```
mrichards42: It would be especially great if we could also log the full list of ocd-ids that the civic info api returned for later analysis, and to make sure that we aren't incorrectly marking any ocd-ids as conflicting.

One way to do that might be changing this function to take the full `result` map instead of `status` and `address`, and then including the civic-info ocd-ids for `:status :conflicting`.

src/user_works/util.clj
```diff
@@ -61,3 +64,31 @@
...
+  A match is when a key and value pair of invalid-pairs is found in ocd-divisions."
+  [ocd-divisions]
+  (let [possible-invalid-matches (map invalid-pairs ocd-divisions)]
+    (not (empty?
+          (filter #(some ocd-divisions %) possible-invalid-matches)))))
+
+(defn wrong-state?
+  "Returns true if the state of a ocd-division does not match the state of the address."
+  [ocd-divisions address-state]
+  (let [division-with-state (first (filter #(re-matches #"ocd-division/country:us/state:\w\w.*" %) ocd-divisions))
```
mrichards42: A few things here:

1. `district:dc` can also be used in place of `state:dc`, so we should account for that case.
2. You should always have a state level ocd-id returned (so you can remove the `.*` at the end)
3. If you use a capture group around the state, you can skip the whole extraction business that happens in the 6 lines below, i.e. `#"ocd-division/country:us/state:(\w\w)"`

src/user_works/util.clj
```diff
@@ -3,6 +3,9 @@
             [clj-time.coerce :as coerce]
             [com.rpl.specter :as specter]))
 
+(def invalid-pairs
+  (hash-map "ocd-division/country:us/state:mi/county:oakland/place:novi" (set ["ocd-division/country:us/state:mi/place:novi"])))
```
mrichards42: 1. It would be customary to use data literals:
```clj
{"ocd-division/country:us/state:mi/county:oakland/place:novi" #{"ocd-division/country:us/state:mi/place:novi"}}
```

2. I would consider making this a list of pairs instead map with sets as values (this would require changing your `invalid-pair?` function, but I think it would make it easier in the long run):
```clj
;; yes technically they're vectors not lists, but 
[["ocd-division/country:us/state:mi/county:oakland/place:novi" "ocd-division/country:us/state:mi/place:novi"]
 ...]
```

3. IMO this belongs in a config file somewhere, but that's not super important.

src/user_works/util.clj
```diff
@@ -61,3 +64,31 @@
   (when minutes
     (-> (time/plus (time/now) (time/minutes minutes))
         coerce/to-date)))
+
+(defn invalid-pair?
+  "Given ocd-divisions, returns true if they contain at least one match to invalid-pairs.
+  A match is when a key and value pair of invalid-pairs is found in ocd-divisions."
+  [ocd-divisions]
+  (let [possible-invalid-matches (map invalid-pairs ocd-divisions)]
+    (not (empty?
```
mrichards42: FYI: `(not (empty?` is the same as [`seq`](https://clojuredocs.org/clojure.core/seq), which is considered idiomatic for checking if a sequence is not empty -- `(empty? x)` is in fact [defined](https://github.com/clojure/clojure/blob/ee3553362de9bc3bfd18d4b0b3381e3483c2a34c/src/clj/clojure/core.clj#L6206-L6211) as `(not (seq x))`

src/user_works/util.clj
```diff
@@ -61,3 +64,31 @@
   (when minutes
     (-> (time/plus (time/now) (time/minutes minutes))
         coerce/to-date)))
+
+(defn invalid-pair?
```
mrichards42: If you were to change the format of `invalid-pairs` to be a list of pairs, one way to do this would be:

```clj
(defn invalid-pair?
  [ocd-divisions]
  (some #(every? ocd-divisions %) invalid-pairs))
```

src/user_works/util.clj
```diff
@@ -61,3 +64,31 @@
...
+        division-state (subs division-with-state state-start-index state-end-index)]
+    (not= division-state (clojure.string/lower-case address-state))))
+
+(defn conflicting?
+  "Given OCD divisions, returns true if there are conflicting OCD division.
+  Conflicting ocd-ids currently include:
+    1. Ocd-divisions where the state doesn't match the address
+    2. Matches to invalid-pair (defined above)."
+  [ocd-divisions address-state]
+  (or (wrong-state? ocd-divisions address-state)
```
dzcode: It might be worth adding some more logging to each of these functions  `wrong-state?` and `invalid-pair?` with  what the mismatch was so we can get some stats on specific issues

src/user_works/util.clj
leilakass: Took care of 2 and 3! Thanks for the regex tip
I had thought about 1 before, but I was thinking that since they are aliased, we should always have `state:dc` when we have `district:dc`. What do you think?

src/user_works/util.clj
leilakass: I had originally this as a invalid list of pairs (or map of pairs), but [after a convo I had with you and Egg ](https://democracyworks.slack.com/archives/C0FUNUSGN/p1583429458059200) I had switched it to map with sets as values. Are you thinking that it would be easier in the long run in terms of adding pairs to the list?

src/user_works/util.clj
mrichards42: You're right that we should always _expect_ to have the state ocd-id, since it's an alias. My comment was more directed to (a) making sure all the bases are covered, and (b) keeping this consistent with other places where we parse ocd-ids (which special-case the `district:dc` case). (e.g. in [election-notification-works](https://github.com/democracyworks/election-notification-works/blob/9aec0fadbbb928a57d4108e2359ad06e3d1cd56d/src/election_notification_works/handlers/util.clj#L231-L235) and [electorate-works](https://github.com/democracyworks/electorate-works/blob/d5a0c9db32c386528e564c79f17a6afc1eca28c2/src/electorate_works/ocdid.clj).

src/user_works/util.clj
mrichards42: Yeah totally, I'm mostly reiterating the [comment](https://democracyworks.slack.com/archives/C0FUNUSGN/p1583432510062800) I wrote in that discussion. I think it'll be easier to maintain if it's a list of pairs.

(or perhaps even better a set of set, since if you try to add a pair to the list that already exists you'll get a compile-time error):
```clj
(def invalid-pairs
  #{#{"ocd-id-1" "ocd-id-2"}
    #{"ocd-id-A" "ocd-id-B"}
    #{"ocd-id-2" "ocd-id-1"}})
;; Syntax error reading source at (...).
;; Duplicate key: #{"ocd-id-2" "ocd-id-1"}
```

src/user_works/util.clj
leilakass: I added some logging but the code isn't so pretty, definitely let me know if there is a better way you are thinking it could work

src/user_works/util.clj
leilakass: Hm is there a reason you think it will be easier to maintain? I think using a map with sets as values is more clearly structured and carries more information than making list of pairs. For example, in the key val pair I just added, we can see clearly that the problem is that the same OCD-ID of ocd-division/country:us/state:in/place:washington is being applied to at least 4 different Washington’s in different counties in IN. I think it's valuable to maintain that information if it's simple enough (which IMO it is in current form).

mrichards42: The main reason I think a list of pairs would be better is that it's a simpler data structure. So (a) IMO it's a little easier to explain, and thus a little easier to understand ("invalid if both of these ocd-ids are present"), and (b) it doesn't assume anything about what sort of pattern there is to invalid ocd-ids.

The the map -> set of invalid ocd-ids is (a) a little trickier ("invalid if this ocd-id is present, and one of these other ocd-ids is present"), and (b) assumes that the typical pattern we'll see is 1 main ocd-id that indicates a group of other ocd-ids are incorrect. I'm not sure we have enough experience w/ invalid ocd-ids to know that for sure, so there may come a point when this structure gets in the way -- e.g. if we need to add an invalid ocd-id for `state:mi/place:novi` (which is in the set under `state:mi/county:oakland/place:novi`), what's the best way to do that?

```clj
;; where we are now
{"ocd-division/country:us/state:mi/county:oakland/place:novi"
 #{"ocd-division/country:us/state:mi/place:novi"}}

;; better to add another hierarchy?
{"ocd-division/country:us/state:mi/county:oakland/place:novi"
 #{"ocd-division/country:us/state:mi/place:novi"}

 "ocd-division/country:us/state:mi/place:novi_township"
 #{"ocd-division/country:us/state:mi/place:novi"}}

;; ... or swap the "master" ocd-id?
{"ocd-division/country:us/state:mi/place:novi"
 #{"ocd-division/country:us/state:mi/county:oakland/place:novi"
   "ocd-division/country:us/state:mi/place:novi_township"}}


;; vs list of pairs: it doesn't matter, you'd just add it to the end of the list
#{#{"ocd-division/country:us/state:mi/county:oakland/place:novi" "ocd-division/country:us/state:mi/place:novi"}
  #{"ocd-division/country:us/state:mi/place:novi" "ocd-division/country:us/state:mi/place:novi_township"}}

```

mrichards42: But this is all bikeshedding at this point. If you think the map -> set option is better, that's totally fine.

src/user_works/conflicting.clj
```diff
@@ -0,0 +1,39 @@
...
+(defn invalid-pair?
+  "Given ocd-divisions, returns true if they contain at least one match to invalid-pairs.
+  A match is when a key and value pair of invalid-pairs is found in ocd-divisions."
+  [ocd-divisions]
+  (let [possible-invalid-matches (map invalid-pairs ocd-divisions)
+        conflicting-id (filter #(some ocd-divisions %) possible-invalid-matches)]
+    (if (boolean (seq conflicting-id))
+      (do
+        (log/debug (str "Conflicting ocd-id found for " (first (first conflicting-id))))
+        true))))
```
mrichards42, line 20: 1. No need for `boolean` here, just `if seq` is fine
2. When you have `if` and no else, it's preferable to use `when` instead (and especially in this case, since you can get rid of the `do`)

    ```clj
    (when (seq conflicting-id)
      (log/debug ...)
      true)
    ```
3. No need for `str` in `log/debug`, you'll get those concatenated for free, though IMO `log/debugf` is even better in most cases (there are some gotchas with using the non `f` version when you're printing a large data structure, so I'm trying to get used to using `f` in all cases).

    ```clj
    (log/debugf "Conflicting ocd-id found for %s" (ffirst conflicting-id))
    ```

4. It would be super helpful if you could log all bad ocd-ids. I think the way this works right now, if you had both `state:mi/place:novi` and `state:mi/county:oakland/place:novi` the log would only include `state:mi/place:novi`, so it wouldn't tell you _why_ that particular ocd-id was invalid.

src/user_works/conflicting.clj
```diff
@@ -0,0 +1,39 @@
...
+
+(defn wrong-state?
+  "Returns true if the state of a ocd-division does not match the state of the address."
+  [ocd-divisions address-state]
+  (let [division-state (last (some #(re-matches #"ocd-division/country:us/state:(\w\w)" %) ocd-divisions))]
+    (if (not= division-state (clojure.string/lower-case address-state))
```
mrichards42, line 26: Nit: it's good practice to `require` all namespaces you need to use (it so happens that this will work, but it won't always). And once you've done usually you'd want to use an alias (`str` or `string` are pretty standard for this library).

src/user_works/conflicting.clj
```diff
@@ -0,0 +1,39 @@
...
+  {"ocd-division/country:us/state:mi/county:oakland/place:novi" #{"ocd-division/country:us/state:mi/place:novi"}
+   "ocd-division/country:us/state:in/place:washington" #{"ocd-division/country:us/state:in/county:marion/place:washington"
+                                                         "ocd-division/country:us/state:in/county:daviess/place:washington"
+                                                         "ocd-division/country:us/state:in/county:tippecanoe/place:washington"
+                                                         "ocd-division/country:us/state:in/county:dearborn/place:washington"}})
+
+(defn invalid-pair?
+  "Given ocd-divisions, returns true if they contain at least one match to invalid-pairs.
+  A match is when a key and value pair of invalid-pairs is found in ocd-divisions."
+  [ocd-divisions]
```
mrichards42, line 14: Nit: since you're using `ocd-divisions` as a set later in this fn, it might be a good idea to make that explicit:

```clj
(let [ocd-divisions (set ocd-divisions) ; <-- this is a no-op if it's already a set
      ...
```

If you were to send in a vector of ocd-ids (like I did while trying out this code 🙂) the `(some ocd-divisions %)` bit throws.

src/user_works/conflicting.clj
```diff
@@ -0,0 +1,39 @@
+(ns user-works.conflicting
```
mrichards42, line 1: Nit: just `conflicting` doesn't make it clear we're talking about ocd-ids . . . IMO it would be better call this ns something like `user-works.ocd-id`, since the full name of the important function would be `user-works.ocd-id/conflicting?` which is self-explanatory to me (where as `user-works.conflicting/conflicting?` isn't as clear).

src/user_works/util.clj
leilakass: Ahhh I see, this makes a lot of sense- updated this!

src/user_works/conflicting.clj
leilakass: thanks for all the comments! really helpful

src/user_works/ocd_id.clj
```diff
@@ -0,0 +1,38 @@
...
+                    #{"ocd-division/country:us/state:in/place:washington" "ocd-division/country:us/state:in/county:daviess/place:washington"}
+                    #{"ocd-division/country:us/state:in/place:washington" "ocd-division/country:us/state:in/county:tippecanoe/place:washington"}
+                    #{"ocd-division/country:us/state:in/place:washington" "ocd-division/country:us/state:in/county:dearborn/place:washington"}])
+
+(defn invalid-pair?
+  "Given a set of ocd-divisions, returns true if they contain at least one match to invalid-pairs.
+  A match is when a key and value pair of invalid-pairs is found in ocd-divisions."
+  [ocd-divisions]
+  (let [ocd-divisions (set ocd-divisions)
+        invalid-pair (some #(when (every? ocd-divisions %) %) invalid-pairs)]
```
mrichards42, line 16: the `(when (every? ...` bit is clever but kind of tricky. I think you would get the same result if you used `filter` instead of `some`, e.g.:
```clj
(->> invalid-pairs
     (filter #(every? ocd-divisions %))
     (first))
```

src/user_works/ocd_id.clj
```diff
@@ -0,0 +1,38 @@
...
+    (when invalid-pair
+      (log/debugf (str "Invalid pair of ocd-ids found: " invalid-pair))
```
mrichards42, line 18: Ah, sorry I wasn't clear about how `debugf` works -- what you have _will_ work, but only accidentally. `log/debugf` (and `infof`, `warnf`, etc.) are supposed to be used with a [printf](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Formatter.html#syntax)-style [`format`](https://clojuredocs.org/clojure.core/format) string
```clj
;; the first arg is a format string
;; other args are used in the format string
(log/debugf "Invalid pair of ocd-ids found: %s" invalid-pair)

;; so, the same as this:
(log/debug (format "Invalid pair od ocd-ids found: %s" invalid-pair))
```

src/user_works/ocd_id.clj
```diff
@@ -0,0 +1,38 @@
...
+      true)))
+
+(defn wrong-state?
+  "Returns true if the state of a ocd-division does not match the state of the address."
+  [ocd-divisions address-state]
+  (let [division-state (or (last (some #(re-matches #"ocd-division/country:us/state:(\w\w)" %) ocd-divisions))
+                           (last (some #(re-matches #"ocd-division/country:us/district:(\w\w)" %) ocd-divisions)))]
+    (when (not= division-state (str/lower-case address-state))
+      (log/debugf (str "State in ocd-division (" (str/upper-case division-state)
+                       ") does not match state in provided address (" (str/upper-case address-state) ")"))
```
mrichards42, line 28: This is really where the format syntax shines, since it makes it clear what the overall message looks like:

```clj
(log/debugf "State in ocd-division (%s) does not match state in provided address (%s)"
            (str/upper-case division-state)
            (str/upper-case address-state))
```

src/user_works/handlers.clj
```diff
@@ -121,6 +123,11 @@
       {:address (assoc address :ocd-divisions fallback-divisions)
        :status :not-found}
 
+      ;; Successful response code, but conflicting OCD IDs found.
+      (ocd-id/conflicting? ocd-divisions address)
+      {:address (assoc address :ocd-divisions fallback-divisions :conflicting-divisions ocd-divisions)
```
mrichards42, line 128: One issue w/ adding `conflicting-division` to the _address_ instead of making a new key in the result map is that we then pass that address around elsewhere in user-works. The other functions expect to see `:ocd-divisions` in an address, but they don't expect `:conflicting-divisions`

... and forgive a little digression, but you could argue that the clojure way is to have those other functions ignore unexpected keys, rather than complain abut them (clojure.spec makes it very easy to declare what keys are _required_, but has no built-in way to restrict which keys are _allowed_, which was done on purpose -- here's some of the rationale, in a rich hickey [talk](https://youtu.be/oyLBGkS5ICk?t=3258)). IMO this is a case where `conflicting-divisions` makes more sense as a property of the result map, rather than the address, but 🤷‍♂ not a huge deal.

src/user_works/ocd_id.clj
leilakass: ah i see! thanks for clarifying!

src/user_works/handlers.clj
leilakass: Added `conflicting-divisions` in the result map!

src/user_works/ocd_id.clj
leilakass: ahh true, thanks! definitely more clear

src/user_works/handlers.clj
```diff
@@ -165,12 +175,12 @@
         address-id (:db/id registered-address)
         user-id (:user/id user)
         current-ocd-divisions (:address/ocd-division-id registered-address)
-        {updated-address :address geocode-status :status}
-        (geocode-address* (util/denamespace-map
+        geocode-result (geocode-address* (util/denamespace-map
                            registered-address
                            address/address-keys))
```
mrichards42, line 180: mind updating the indentation of these two lines, since the line above it got changed?

test/user_works/ocd_id_test.clj
```diff
@@ -0,0 +1,28 @@
+(ns user-works.ocd-id-test
+  (:require
+   [clojure.test :refer :all]
+   [user-works.ocd-id :refer :all]))
```
mrichards42, line 4: A lot of our test code looks like this, but as a rule we try to avoid `:refer :all`. Usually what that means is:

```clj
(:require [clojure.test :refer [deftest testing is]]
          [user-works.ocd-id :as ocd-id])

;; and then

(deftest wrong-state-test
  (let [va-divs ...]
    (testing "true if wrong state"
      (is (true? (ocd-id/wrong-state? va-divs "MD"))))
    (testing "false if correct state"
      (is (nil? (ocd-id/wrong-state? va-divs "VA"))))))
```

test/user_works/ocd_id_test.clj
```diff
@@ -0,0 +1,28 @@
...
+(deftest invalid-pairs-test
+  (let [invalid-divs #{"ocd-division/country:us/state:in/county:marion/place:washington"
+                       "ocd-division/country:us"
+                       "ocd-division/country:us/state:in/place:washington"}
+        valid-divs #{"ocd-division/country:us"
+                     "ocd-division/country:us/state:in/place:washington"}]
+    (testing "true if conflicting pairs"
+      (is (true? (invalid-pair? invalid-divs))))
+    (testing "false if only has one match within a pair"
+      (is (nil? (invalid-pair? valid-divs))))))
```
mrichards42, line 28: The return type is important, so we probably shouldn't test the type explicitly (or if the return type _is_ important, we should document it).

Probably simplest to do:
```clj
(is (invalid-pair? invalid-divs))
(is (not (invalid-pair? valid-divs)))

;; and in the other test
(is (wrong-state? va-divs "MD"))
(is (not (wrong-state? va-divs "VA")))
```

PR #157, opened 2020-04-06
Tvi 40 user ready timing

solomonsokoya:
Added a new queue to `user-works` which will help changes the welcome noty delays based on where the user is during the signup phase. 

[Jira Card](https://democracyworks.atlassian.net/browse/TVI-1?atlOrigin=eyJpIjoiNzFiNGE2ZGI4NTU1NGNkZTg0ZDA3MzhmYTc0MTdjOWMiLCJwIjoiaiJ9)


resources/config.edn
```diff
@@ -81,7 +81,13 @@
...
                                 :response true
                                 :prefetch-limit #resource-config/edn #resource-config/env "USER_WORKS_REPORTING_QUERY_PREFETCH"
-                                :threads #resource-config/edn #resource-config/env "USER_WORKS_REPORTING_QUERY_THREADS"}]
+                                :threads #resource-config/edn #resource-config/env "USER_WORKS_REPORTING_QUERY_THREADS"}
+                               {:queue "user-works.user.update-user-ready-timing"
+                                :f user-works.handlers/update-ready-timing
+                                :response true
+                                :prefetch-limit #resource-config/edn #resource-config/env "USER_WORKS_USER_UPDATE_EVENT_PREFETCH"
+                                :threads #resource-config/edn #resource-config/env "USER_WORKS_USER_UPDATE_EVENT_THREADS"}
+                               ]
```
sduckett, line 90: nit: this dangling `]` could be lifted next to the closing `}`

resources/config.edn
```diff
@@ -109,4 +115,9 @@
  {:uri #resource-config/env "USER_WORKS_DATOMIC_URI"
   :partition :user-works}
  :civic-info
- {:api-key #resource-config/env "GOOGLE_CIVIC_API_AUTH_KEY"}}
+ {:api-key #resource-config/env "GOOGLE_CIVIC_API_AUTH_KEY"}
+ :user-ready-timing
+ {:default 7200000
```
sduckett, line 120: 🤔 I don't have a good suggestion here, but I wish we could express this differently: instead of `7200000`, we could write something like `[2 :hours]` and avoid doing math when reading/writing config files.

src/user_works/handlers.clj
```diff
@@ -218,6 +222,11 @@
   {:phone s/Str
    (s/optional-key :email) s/Str})
 
+(def timing-schema
```
mrichards42, line 225: Looks like this is unused?

src/user_works/handlers.clj
```diff
@@ -26,12 +26,16 @@
 (defstate google-api-client
   :start (civic-info/google-api-client (rc/config [:civic-info :api-key])))
 
+(defn timing [action-name]
+  (if (nil? action-name)
```
sduckett, line 30: If `action-name` is non-nil, but not present in the config file, this will throw an exception.

src/user_works/handlers.clj
```diff
@@ -405,6 +432,11 @@
       util/strip-namespaces-from-message
       handle-errors))
 
+(def update-ready-timing
```
mrichards42, line 435: Nit: usually we put these wrapper fns right under the one they're wrapping (so, right about line 407)

src/user_works/handlers.clj
```diff
@@ -378,6 +387,24 @@
    (merge {:id s/Uuid} user-schema email-schema)
    (merge {:id s/Uuid} user-schema phone-schema)))
 
+(def update-ready-timing-schema
+  {:id s/Uuid
+   :event-timing s/Keyword})
+
+(defn-traced update-ready-timing*
+  [{:keys [id event-timing] :as user-map}]
+  (log/debug "Received a user ready event timing update message:" (pr-str user-map))
```
sduckett, line 396: There's a helper for this:
```suggestion
  (log/debugf "Received a user ready event timing update message: %s" user-map)
```

src/user_works/handlers.clj
```diff
@@ -378,6 +387,24 @@
...
+  (s/validate update-ready-timing-schema user-map)
+  (if-let [user (user/find-by-id (db/db) id)]
```
mrichards42, line 398: An optimization: you don't need to fetch the user here, since scheduled-message-works allows you to reschedule messages [without overwriting their content](https://github.com/democracyworks/scheduled-message-works/blob/8ae97f68d5baed611dceabe15264a2b322db708e/src/scheduled_message_works/db/sql.clj#L90).

The original message is required to have a `:message` key:

```clj
{:send-to "user.ready"
 :send-after (util/minutes-from-now delay)
 :identifiers {:user-id user-id
               :notification-type :user-ready}
 :message {:user-id user-id
           :hostname hostname}}
```

... but you can reschedule without a message:

```clj
{:send-to "user.ready"
 :send-after (util/minutes-from-now delay)
 :identifiers {:user-id user-id
               :notification-type :user-ready}}
```

src/user_works/handlers.clj
```diff
@@ -26,12 +26,16 @@
 (defstate google-api-client
   :start (civic-info/google-api-client (rc/config [:civic-info :api-key])))
 
+(defn timing [action-name]
+  (if (nil? action-name)
+    (rc/config [:user-ready-timing :default])
+    (rc/config [:user-ready-timing action-name])))
```
mrichards42, line 32: Rather than `:default`, I think it would be clearer to use `:create`, and require that `action-name` be present in the config. Then you'd pass `{:action-name :create}` in the map to `schedule-user-ready-event` in the user-create handler.

src/user_works/handlers.clj
solomonsokoya: Wouldn't we want to check if the user exist?

solomonsokoya: the `reschedule-user-ready-event` and `schedule-user-ready-event` use the same fn `schedule-user-ready-event*` are you implying not to pass `:hostname`?

resources/config.edn
solomonsokoya: I wanted to keep in consistent throughout the app.  Are you suggesting it should be a comment?

sduckett: Yes, please: A comment would be very helpful here.

For now, I'm going to let my thought about expressiveness marinade in the back of my mind.

src/user_works/handlers.clj
mrichards42: > Wouldn't we want to check if the user exist?

It shouldn't matter -- rescheduling a message only has an effect if there's already a scheduled message. And if we have a scheduled `user.ready` message, it's because a user has been created at some point.

And if you send a junk user-id, it won't match any scheduled messages, and nothing will happen.

mrichards42: > the reschedule-user-ready-event and schedule-user-ready-event use the same fn schedule-user-ready-event* are you implying not to pass :hostname?

We'd probably want two functions:
```clj
;; not necessarily these exact args, but this is the behavior we'd want
(defn schedule-user-ready-event [{:keys [user-id hostname]} delay]
  ;; schedule with `:message`
  )

(defn reschedule-user-ready-event [{:keys [user-id]} delay]
  ;; reschedule without `:message`
  )
```

solomonsokoya: Maybe this instead? lots of similarity with the two fn's 
```
(if (= action-name :create)
                             :message {:user-id user-id :hostname hostname})
```

mrichards42: Sounds reasonable to me!

resources/config.edn
```diff
@@ -109,4 +113,9 @@
  {:uri #resource-config/env "USER_WORKS_DATOMIC_URI"
   :partition :user-works}
  :civic-info
- {:api-key #resource-config/env "GOOGLE_CIVIC_API_AUTH_KEY"}}
+ {:api-key #resource-config/env "GOOGLE_CIVIC_API_AUTH_KEY"}
+ :user-ready-timing
+ {:create 7200000 ; = 2hrs
```
mrichards42, line 118: Mind sticking a comment in for each of these actually? All the 0s make the others not super obvious also

src/user_works/handlers.clj
```diff
@@ -392,7 +419,8 @@
       (async/>!! events-chan {:before (user/entity->user user)
                               :after updated-user})
       (reschedule-user-ready-event {:user-id id
-                                    :hostname (:hostname updated-user)})
+                                    :hostname (:hostname updated-user)
+                                    :action-name :create})
```
mrichards42, line 423: We actually want to delete this call entirely in the user-update handler, and just do the create timing in user-create

src/user_works/handlers.clj
```diff
@@ -26,17 +26,25 @@
 (defstate google-api-client
   :start (civic-info/google-api-client (rc/config [:civic-info :api-key])))
 
+(defn timing [action-name]
+    (rc/config [:user-ready-timing action-name]))
```
mrichards42, line 30: Nit: I might call this `user-ready-delay` since (a) we're talking about a delay from now, and (b) just "timing" doesn't tell me what kind of timing we're talking about.

src/user_works/handlers.clj
```diff
@@ -378,6 +387,24 @@
...
+   :action s/Keyword})
+
+(defn-traced user-signup-event*
+  [{:keys [id action] :as user-map}]
+  (log/debugf "Received a user ready event timing update message: %s" user-map)
+  (s/validate user-signup-event-schema user-map)
+  (do (reschedule-user-ready-event {:user-id id
+                                    :action-name action})
+      {:status :ok
+       :user user-map}))
```
mrichards42, line 401: Since this is now a kehaar event handler, and event handlers don't return a response, you can just make this
```clj
(reschedule-user-ready-event {:user-id id :action-name action})
```

src/user_works/handlers.clj
```diff
@@ -26,17 +26,25 @@
...
+    {:message
+     {:user-id user-id
+      :hostname hostname}}))
+
 (defn-traced schedule-user-ready-event*
-  "Given a :user-id, schedules ready notifications to go in 5 minutes,
-  depending on user updates."
-  [schedule-fn {:keys [user-id hostname]}]
+  "Given a :user-id and action, schedules ready notifications to go based on keyword/default value"
+  [schedule-fn {:keys [user-id hostname action-name]}]
```
mrichards42, line 40: Nit: you're calling this `action` elsewhere (e.g. in the schema, in http-api) so I'd go with `action` everywhere instead of `action-name` in some places.

Or perhaps `signup-action` in this function, since you don't necessarily have the full context that this is in response to a signup action (whereas in the schema and http-api it's clear we're talking about handling a signup event).

src/user_works/handlers.clj
```diff
@@ -378,6 +387,24 @@
    (merge {:id s/Uuid} user-schema email-schema)
    (merge {:id s/Uuid} user-schema phone-schema)))
 
+(def user-signup-event-schema
+  {:id s/Uuid
+   :action s/Keyword})
+
+(defn-traced user-signup-event*
+  [{:keys [id action] :as user-map}]
```
mrichards42, line 395: Nit: usually we'd call this `message` since we're handling a rabbitmq message. I see that it's called `user-map` in the update-user handler, but (a) I suppose technically it _is_ a full user that's being passed in the update message, and (b) I still would have called it `message` in the user-update handler for consistency 🙂

